PR #64, opened 2020-11-05
Minor fixes + Copy editing

drkameleon:


jfecher:
Ante's C++ compiler will be abandoned shortly for the compiler in ante-rs.
These edits can go through but they'll likely be overwritten when that changeover occurs

Issue #67, opened 2021-01-21
Change the lambda syntax from `\a b.y` to `fn a b -> y`


jfecher:
This should be a relatively simple aesthetic change. Should only require changing the expected tokens for the `lambda` rule in the `parser/mod.rs` and adding the `fn` token to the lexer.

This is just a syntax change, so the current lambda `\a b. body` should be exactly equivalent to `fn a b -> body` in the new syntax.

Any places in the `examples` test directory that use lambdas will also need to have their syntax updated.

jfecher:
Changing `\` to `fn` is now done in #70, the remaining work is to change `.` to `->` which likely requires changing the return-type syntax which currently uses `->` as these would conflict: #68.

PR #70, opened 2021-04-18
changed \ to fn for lambda declaration

mkhan45:
This is a really clean codebase!

A few things that would make contributing easier:
- Document how the goldentests work
- Add cargo fmt rules to a rustfmt.toml or just disable it completely (just add `ignore = ["/"]` to rustfmt.toml at the repo root)


jfecher (approved):
Changes look good :)

I agree with your points on making contributing easier, I think I'll make a Contributing wiki page so that I can expand that section with information on goldentests and possibly a FAQ in the future without cluttering the README too much.

I'll explore some cargo formatting rules too, I'm still stuck in the mindset of manually formatting things but even if I don't mind some formatting inconsistencies it does seem easier for contributors in the future to contribute with the ease of mind their formatting is acceptable.

jfecher:
I went ahead and updated #67 too

PR #72, opened 2021-04-28
Desugar parsing for <| and |>

### LINKED ISSUE #65: Desugar parsing for <| and |> directly into function calls
The tokens for these operators already exist: `Token::ApplyLeft` and `Token::ApplyRight` respectively.

`|>` should be left-associative with a precedence two levels under `+`
`<|` should be right-associative with a precedence one level under `+`

Support for these can be added in the parser such that
`a |> foo b c |> bar d e` parses as `bar (foo a b c) d e`
and
`bar d e <| foo b c <| a` also parses as `bar (foo a b c) d e`.

It would be nice if the parser desugared the operator directly into the above function calls so no further handling would need to be done in later passes.

thehabbos007:
Closes #65 

This is still a little incomplete, as I am not quite getting what I expect in the test. I could use some pointers :)

Are the new test correct? I would expect the result to be `(foo ('+' ('+' 1 2) 3) ('+' a b))`  in the case of `1 + 2 + 3 |> foo a + b` with the given precedence change.

Also I don't quite like the way i am dealing with the args vec, any thoughts on that?

Thanks for taking a look! :)


src/parser/mod.rs
```diff
@@ -397,11 +397,44 @@ fn pop_operator<'c>(operator_stack: &mut Vec<&Token>, results: &mut Vec<(Ast<'c>
...
+    let operator = operator_stack.pop().unwrap().clone();
+    let call =  desugar_apply_operator(operator, lhs, rhs, location);
     results.push((call, location));
 }
 
+fn desugar_apply_operator<'a>(operator: Token, lhs: Ast<'a>, rhs: Ast<'a>, location: Location<'a>) -> Ast<'a> {
+    match operator {
+        Token::ApplyLeft => 
+            if let Ast::FunctionCall(fun) = lhs {
+                let mut args = vec![rhs];
```
thehabbos007, line 409: Arg vec handling mentioned

src/parser/mod.rs
```diff
@@ -367,14 +367,14 @@ fn precedence(token: &Token) -> Option<(i8, bool)> {
...
-        Token::Append => Some((9, false)),
-        Token::Range => Some((10, false)),
+        Token::Or => Some((2, false)),
+        Token::And => Some((3, false)),
+        Token::EqualEqual | Token::Is | Token::Isnt | Token::NotEqual | Token::GreaterThan | Token::LessThan | Token::GreaterThanOrEqual | Token::LessThanOrEqual => Some((5, false)),
+        Token::In => Some((6, false)),
+        Token::Append => Some((7, false)),
+        Token::Range => Some((8, false)),
+        Token::ApplyRight => Some((9, false)),
+        Token::ApplyLeft => Some((10, true)),
```
jfecher, line 377: I'm rethinking this - I think ApplyLeft/Right should actually be pretty low in precedence. Possibly lower than Token::Comma.
Would you expect `1, 2 |> print` to print out `1, 2` or just the `2`?

Looking at OCaml for inspiration, OCaml actually has `|>` just above Token::And. That'd allow for `true and false |> not` which seems a bit odd to me.

I think I'm leaning towards it being below Token::Comma with `1, 2 |> print` printing out `1, 2`.

src/parser/mod.rs
jfecher: I think something like the below would work (untested):
```rust
fn desugar_apply_operator<'a>(operator: Token, lhs: Ast<'a>, rhs: Ast<'a>, location: Location<'a>) -> Ast<'a> {
    match operator {
        Token::ApplyLeft => add_argument_to_function(lhs, rhs, location),
        Token::ApplyRight => add_argument_to_function(rhs, lhs, location),
        _ => {
            let operator = Ast::operator(operator, location);
            Ast::function_call(operator, vec![lhs, rhs], location)
        }
    }
}

fn add_argument_to_function<'a>(f: Ast<'a>, arg: Ast<'a>, location: Location<'a>) -> Ast<'a> {
    if let Ast::FunctionCall(mut call) = f {
        call.args.insert(0, arg);
        Ast::FunctionCall(call)
    } else {
        Ast::function_call(f, vec![arg], location)
    }
}
```

jfecher: Maybe that function could also be called `prepend_argument_to_function` for extra clarity.

src/parser/mod.rs
thehabbos007: I would be more used to the `1, 2 |> print` printing `2` than `1, 2`. I believe that would be the case in OCaml as well.
I haven't used piping in tuples that much, though, so I don't have a strong opinion about it

src/parser/mod.rs
thehabbos007: Thank you, missed the `mut call` so I thought I had to allocate a new vec

examples/parsing/apply_operators.an
```diff
@@ -0,0 +1,17 @@
...
+1 + 2 + 3 |> foo a + b
+foo a + b <| 1 + 2 + 3
+// args: --parse
+// expected stdout:
+// (foo a);
+// (foo b 2);
+// (bar (foo a b c) d e);
+// (bar (foo a b c) d e);
+// (foo ('+' ('+' 1 2) 3) ('+' a b));
+// (foo ('+' ('+' 1 2) 3) ('+' a b))
```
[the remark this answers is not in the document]
thehabbos007: Ah my bad. This makes sense, I'll try to find a better test case for the precedense after we figure out where we want it in the precedence list :)

src/parser/mod.rs
jfecher: I did a quick survey on the discord, looks like most expect it to output `1, 2`. I think lets just go with that, if it ends up being confusing it is easy to change later

examples/parsing/apply_operators.an
```diff
@@ -5,13 +5,17 @@ b |> foo 2
 a |> foo b c |> bar d e
 bar d e <| foo b c <| a
 
-1 + 2 + 3 |> foo a + b
-foo a + b <| 1 + 2 + 3
+1, 2 |> baz
+baz <| 1, 2
+
+1, 2 |> baz <| 3, 4
```
jfecher, line 11: Good testcase :+1:

jfecher (approved):
Looks good, thanks for getting this done!

Looks like I'll need to add some more to the quick tasks list real soon - alternatively there's also the tour page on the website which is the closest thing ante has to a full spec if you're looking for more things to do. :+1:

thehabbos007:
Ah cool, thank you for making it easy to contribute! :)

PR #73, opened 2021-05-02
Explicit curry

### LINKED ISSUE #66: Add support for explicit currying via `_`
When used in a function-call, the `_` marks an explicitly curried parameter since ante does not have implicit currying.

The following:
```hs
foo a _ c
bar _ _ 3
```
Should be equivalent to (_param names are unimportant/placeholders):
```hs
fn _param1 -> foo a _param1 c
fn _param1 _param2 -> bar _param1 _param2 3
```

This should be a relatively easy check to do manually after parsing a `Ast::FunctionCall` in `parser/mod.rs`. After a `_` is found in a function call, the parser can manually desugar it by returning an `Ast::Lambda` containing the function call.

thehabbos007:
Closes #66 

I went with new allocations, as I was having trouble with a mutable version. But I think the extra allocations are negligible.

Have commented a couple places where there might be better options :)


src/parser/mod.rs
```diff
@@ -471,9 +471,47 @@ fn term<'a, 'b>(input: Input<'a, 'b>) -> AstResult<'a, 'b> {
...
+
+    let function_call = Ast::function_call(function, args, loc);
+    Ast::lambda(curried_args, None, function_call, loc)
+}
+
+fn matches_underscore(arg: &Ast) -> bool {
+    matches!(arg, Ast::Variable(ast::Variable{ kind: ast::VariableKind::Identifier(x), ..}) if x == "_")
+}
+
+fn matches_not_typeconstructor(function: &Ast) -> bool {
```
thehabbos007, line 511: Is this and the other `matches_underscore` function implemented somewhere else or is this alright? We could put it in an impl block by the enum definitions, but it might be a little overkill just for one place :)

src/parser/mod.rs
```diff
@@ -471,9 +471,47 @@ fn term<'a, 'b>(input: Input<'a, 'b>) -> AstResult<'a, 'b> {
...
+                let curried_arg = format!("_${}", curried_arg_count);
+                curried_args.push(Ast::variable(curried_arg.clone(), loc));
+                Ast::variable(curried_arg, loc)
+            } else {
+                arg
+            }
+        })
+        .collect();
+
+    let function_call = Ast::function_call(function, args, loc);
```
thehabbos007, line 503: I am using the same loc here everywhere, that feels kind of incorrect, but am unsure if it is

examples/parsing/explicit_curry.an
```diff
@@ -0,0 +1,17 @@
...
+
+bar _ "test" _
+
+baz 1 (_, 32)
+qux 3 (_) _
+
+
+// args: --parse
+// expected stdout:
+// (\ _$1 . (foo _$1 a));
```
jfecher, line 13: Ah, we should probably update the ast printout syntax for lambdas in a future PR to match the new `fn params -> body` syntax

examples/parsing/explicit_curry.an
```diff
@@ -0,0 +1,17 @@
...
+// (\ _$1 . (oof 1 _$1 3));
+// (\ _$1 _$2 . (bar _$1 "test" _$2));
+// (baz 1 (',' _ 32));
```
jfecher, line 16: I'd expect `(',' _ 32)` to be a lambda here - `(\ $1 . (',' $1 32))` (see comment below)

src/parser/mod.rs
jfecher: I think there's precedence for being in an impl block though (there are some existing functions there like `is_pair_constructor` I believe). Since these are only used for parsing I'm fine with them either staying here or being in an impl.

I don't think we need a `matches_not_typeconstructor` check though. Ideally we should be able to curry type constructors too. Something like the following should work:
```hs
type Person = name: string, job: string

names = ["alice", "bob", ...]
people = map names (Person _ "programmer")
```
Conceptually, type constructors are just functions so there's no need to special case them here. For example the `Person` type constructor has type `string string -> Person`

src/parser/mod.rs
jfecher: Using the `arg`'s location for creating the variables on lines 495 and 496 would be nice. You can get that with `arg.locate()` which requires the Locatable trait which I think is already imported

src/parser/mod.rs
```diff
@@ -471,9 +471,47 @@ fn term<'a, 'b>(input: Input<'a, 'b>) -> AstResult<'a, 'b> {
...
+    }
+
+    Ast::function_call(function, args, loc)
+}
+
+fn curried_function_call<'a>(function: Ast<'a>, args: Vec<Ast<'a>>, loc: Location<'a>) -> Ast<'a> {
+    let mut curried_args = vec![];
+    let mut curried_arg_count = 0;
+    let args: Vec<Ast<'a>> = args
+        .into_iter()
```
jfecher, line 490: This `into_iter().map(...).collect()` pattern is essentially what `fmap` does in the utils module but fmap only takes references. Maybe we could change it to take a generic sized iterator instead. This doesn't have to be in this PR though

src/parser/mod.rs
```diff
@@ -471,9 +471,47 @@ fn term<'a, 'b>(input: Input<'a, 'b>) -> AstResult<'a, 'b> {
...
+        .map(|arg| {
+            if matches_underscore(&arg) {
+                curried_arg_count += 1;
+                let curried_arg = format!("_${}", curried_arg_count);
```
jfecher, line 494: Nitpick: Can we get rid of the `_` in front of the generated names? That way it'd match the `$1, $2, ..., $n` syntax used by some other languages with their lambda syntaxes.

examples/parsing/explicit_curry.an
jfecher: Also, I hate to be ante-fun but can we change oof's name to something else? Following foo, bar, baz, `qux` usually comes after those three

thehabbos007: I had written qux originally, but I dont remember why i changed it. Will change it back :)

src/parser/mod.rs
thehabbos007: I didn't have the check at first, but the desugaring logic triggered on match arms for match statements and resulted in
```
   | examples/typechecking/completeness_checking.an: 7,4        error: Type mismatch between (Maybe a) and (a -> (Maybe a))
   | | (_, Some _) -> ()
```
What should we do about this situation?

src/parser/mod.rs
thehabbos007: Ah yes, makes sense as it is a common pattern for the vecs. I could look into that after fixing up some of the lambda pretty prints!

examples/parsing/explicit_curry.an
thehabbos007: > 
> 
> I'd expect `(',' _ 32)` to be a lambda here - `(\ $1 . (',' $1 32))` (see comment below)
But the curried arg shouldn't bubble up to the sorrounding lambda, right? And should the explicit currying happen at arbitrary depths?

src/parser/mod.rs
jfecher: Oh, looks like we need to not desugar if we're in a pattern match. This could complicate the parser some. Changing `irrefutable_pattern` would be easy but refutable patterns just re-use the `expr` parser which makes it harder. If you find a good way of including it then we can add that in but otherwise I'm fine with leaving this as a TODO for later.

examples/parsing/explicit_curry.an
jfecher: Correct, it shouldn't bubble up it should only happen for the function call its used in which in this case looks to be `,`. So the second argument of `baz` should be this function

jfecher: Since `,` is a type constructor this case may just not be desugared yet because of the temporary type constructor check

jfecher (approved):
Looks like the only remaining items are TODOs that can be done in other PRs, good work :+1: 

In addition to the work above another task could be to make sure this new `function _ arg` syntax works with the apply operators so you can still chain functions even when you don't want to use the first argument:

```hs
second a b = b

map array (_ + 1)
|> filter fn x -> x > 3
|> second 0 _
```
I think the largest priorities are getting explicit currying working for type constructors then getting it working with the apply operators. Another task you could do is creating a primitive `Ptr` type which is like `ref` but without lifetime inference. This could be fun since it'd require changes at each pass of the compiler. I could stand to write up some new issues now that the old ones are all complete (thanks!).

thehabbos007:
Thanks for letting me have a go at the issues, has been real fun! Would be great to tackle some more. The Ptr one sounds a little difficult, but I'm up for a challenge if you believe I am able to do it :)

Issue #76, opened 2021-05-02
Explicit currying for type constructors


jfecher:
Since type constructors are just functions internally, they should also work with the explicit currying syntax. Currently this is not the case due to a syntax conflict when using `_` to throw away variable names in match patterns:
```scala
match foo with
| Bar _ b -> b
...
```
The `Bar _ b` here out of context has the same syntax as an explicitly curried function but a different meaning since when used in a pattern it is a deconstruction. Ideally after this change, this would remain a deconstruction and we would only explicitly curry functions when they are not used within a pattern.

## Estimated work
- Differentiating between patterns and exprs within the parser means that match patterns can likely no longer reuse the `expr` parser. They'll likely need their own `refutable_pattern` parser which can hopefully be written using the existing `irrefutable_pattern` parser plus some other things like function calls.
- In the future (or perhaps in this change) its possible we may merge both into a single `pattern` parser since only name resolution and on will be able to tell if some pattern `(Foo a b)` is refutable or not (based on whether the type Foo belongs to has only 1 or more than 1 constructor). So another reasonable choice in this PR is to merge everything into a single `pattern` parser.
- Explicit currying should then only be within the `expr` parser and not the `pattern` parser.
- This is expected to break some tests for explicit currying, these will need to be updated as well.

jfecher:
Ended up creating the separate `pattern` parser which match patterns now use.

PR #81, opened 2021-05-17
Apply operators with explicit currying

### LINKED ISSUE #75: Apply operators should work with explicit currying
The apply operators `<|` and `|>` should still work when their function parameter is an explicitly curried function. These operators always currently apply their argument to the first argument of the function - with explicit currying the argument should be slotted in wherever the `_` is in the function.

```hs
second a b = b

two = 3 |> second 2
too = 3 |> second _ 2
three = 3 |> second 2 _

four = second 4 <| 5
fore = second _ 4 <| 5
five = second 4 _ <| 5
```

## Estimated work
- The core desugaring may actually already work for this usecase - I'm not sure. If not the apply operators may need a check for explicit currying of their function argument or to check if their argument is already a lambda.
- There should be new parsing tests for this functionality

thehabbos007:
Closes #75 

I've taken the opportunity to also put the desugaring functions in another file, as it was getting a little long.

The slotting works with literals and variables, as I believe there might be unintended consequences if I were to slot for example an impure function call slotting into `fn x -> x + x`. This is both for the sake of not putting clone on everything on the AST, and for uninteded execution.

There could certainly be some improvements, but am unsure what that could be. Will change up the stuff you see can be simplified :)


src/parser/desugar.rs
```diff
@@ -0,0 +1,116 @@
...
+fn replace_lambda_body<'a>(f: &mut ast::FunctionCall<'a>, new_arg: Ast<'a>, arg_replace: Ast<'a>) {
+    let idxs: Vec<usize> = f.args
+                .iter()
+                .enumerate()
+                .filter(|(_, x)| name_matches(&arg_replace, x))
+                .map(|(i, _)| i).collect();
+
+    for idx in idxs {
+        match &new_arg {
+            Ast::Literal(lit) => {
```
jfecher, line 99: Honestly, I'm having second thoughts with this optimization. Applying it to all lambdas complicates things somewhat and makes me think that this is maybe better left to the optimizer. I thought it might be salvagable by changing the check to see if the argument is only used once instead, but that could still change evaluation order of something like `(print 1; 5) |> foo (print 2; 10) _`.
I think it'd be simpler if we just kept it as immediately calling the lambda so `3 |> print _` = `(fn $1 -> print $1) 3`

thehabbos007: I do agree, I think the potential savings down the line in terms of time to optimize isn't worth the complexity. What if i roll back the slotting and keep the seperate `desugar.rs` file?

jfecher (approved):
Looks good, thanks for the exploration work.

Do we have any codegen tests for this?

thehabbos007:
Good point, I don't think we have any. Let me add some real quick!

jfecher:
Ah, I see it now. Haven't merged my local branch with it yet. Lets add one more test to it like `4 |> (_ + 2) |> print` and call it for this PR.

(actually, on second thought, does `_` work for operators yet? If not you can replace the above test with a different function call)

thehabbos007:
I think I can look into that now, might as well :)

thehabbos007:
There was an easy fix for it, but it needs to be aware if its inside a match, otherwise the same problems are the typeconstructor occur. For now I've just added the new test using (+).

When  #76 is fixed, explicit currying for inline operators could be added in without very cumbersome checks :)

jfecher:
Thanks! The prefix operator in the test is another good one, I don't think I had any tests for operators used in a prefix position at all before

PR #86, opened 2021-06-08
Initial Ptr type addition

### LINKED ISSUE #78: Add primitive Ptr type
Currently, ante's only pointer type is `ref` which incurs lifetime inference semantics that aren't always wanted for other pointer types being built on top of it. Since ante is a low-level language it needs a raw pointer type on top of which other types like `Box t` and `Rc t` can be built on top of. Additionally, it should be useful for interfacing with `extern` functions, which currently incorrectly use `ref`.

This PR should include the addition of the `Ptr a` type and changes to the stdlib to update the extern C functions to take or return pointers when possible. The standard `string` type should also probably be
`type string = len: usz, c_string: Ptr char` instead of `type string = len: usz, c_string: ref char`. This is defined in `builtins.rs`.

There doesn't need to be any constructors or anything added to the Ptr type for now, changing the types of `malloc` and `free` in the stdlib should be sufficient to work with them:
```hs
extern
    malloc: usz -> Ptr a
    free: (Ptr a) -> i32
```

## Estimated work
- This sounds more daunting initially than it likely is.
- Adding a `Ptr` to the `Type` union (possibly in `PrimitiveType`?) should give a bunch of compiler errors for incomplete `match`s which should be all the locations that need to be changed within the compiler for this PR.
- The behavior of `Ptr` within each match can almost be copied 1-to-1 with whatever the `ref` type does within that match with anything related to lifetimes removed.
- Since `Ptr` is not a keyword, `Ptr` may have to be defined as a builtin like `string` is in `src/nameresolution/builtins.rs`. `Ptr` shouldn't be as commonly used as `string` though so we may want to in the future move it to another module so users have to explicitly import it. Unfortunately there's no mechanism for defining a builtin type that isn't always in scope yet. Adding this is outside the scope of this PR.

thehabbos007:
Closes #78 

Currently I am not using `Ptr` for the built-in string type, as I had the following problems when switching out the type in the string struct (with some debug prints)
```
142| [src/types/typechecker.rs:470] &a = UserDefinedType(
143|     TypeInfoId(
144|         2,
145|     ),
146| )
147| [src/types/typechecker.rs:471] &b = Primitive(
148|     Ptr,
149| )
150| /home/ahmad/.config/ante/stdlib/prelude.an: 189,41 error: Type mismatch between Ptr and Ptr
151| impl Print i32 with printne x = printf ("%d".c_string) x
```
I suspect this has to do with the fact that I was using the primitive type instead of the scoped Ptr type here
https://github.com/jfecher/ante/blob/master/src/nameresolution/builtin.rs#L89
And even when referring to the `TypeInfoId`, I reached unreachable code unfortunately.

I think changing string (as well as other places where `Ptr` should replace `ref`) should be in another issue :)

I did change `malloc` and `free` to take `Ptr` instead, though. Any other of those externs need the same treatment? I suspect `realloc` and such should have this, but just want to makre sure.

Do you want some tests for Ptr? And if so how would we test this? :)


src/nameresolution/builtin.rs
```diff
@@ -70,8 +77,18 @@ pub fn import_prelude<'a>(resolver: &mut NameResolver, cache: &mut ModuleCache<'
 
     // Manually insert some builtins as if they were defined in the prelude
     resolver.current_scope().traits.insert("Int".into(), cache.int_trait);
+
+    // Add string to scope
+    resolver.current_scope().types.insert("string".to_string(), STRING_TYPE);
+    resolver.current_scope().definitions.insert("string".to_string(), BUILTIN_ID);
```
thehabbos007, line 83: Is this what you wanted for the string type?

src/types/mod.rs
```diff
@@ -38,6 +38,7 @@ pub enum PrimitiveType {
     CharType,                 // : *
     BooleanType,              // : *
     UnitType,                 // : *
+    Ptr,                      // : * -> *
```
thehabbos007, line 41: I saw in an earlier commit that `ref` in primitive had this comment. Is this correct?

src/nameresolution/builtin.rs
jfecher: Almost - `BUILTIN_ID` defines the `builtin` symbol which is used in the prelude to define some builtin things like integer addition - so it isn't the write ID for the string constructor. Looks like we don't have a constant for the string constructor id yet - so it'll need to be added like Pair's is.
I also don't think the `string` type itself needs to be imported since it has a keyword that ante knows about (there's a note about this above `define_string`) but as long as its not causing tests to fail it may be fine.

src/nameresolution/builtin.rs
```diff
@@ -103,6 +120,35 @@ fn define_string<'a>(cache: &mut ModuleCache<'a>) -> Type {
...
+/// type Ptr a = Ptr a
+fn define_ptr<'a>(cache: &mut ModuleCache<'a>) {
+    let location = Location::builtin();
+
+    let a = cache.next_type_variable_id(LetBindingLevel(0));
+    let ptr = Type::Primitive(PrimitiveType::Ptr);
+    
+    let name = "Ptr".to_owned();
+    let ptr_id = cache.push_type_info(name.clone(), vec![a], location);
+    assert_eq!(ptr_id, PTR_TYPE);
```
jfecher, line 134: Looks like this is the source of your type errors - you define a new `Ptr` type here which is distinct from the `PrimitiveType::Ptr` type. You should hopefully be able to remove these lines and the PTR_TYPE constant (or change `PTR_TYPE` to equal `Type::Primitive(PrimitiveType::Ptr)`)

src/types/mod.rs
jfecher: Yep! `* -> *` is the _kind_ of `Ptr`. In kind terminology `*` means type so `* -> *` means `Ptr` is a function which takes a type and returns a type, so you can give it `i32` as an argument: `Ptr i32` and you get back the type of a pointer to an integer.
Kinds aren't really checked at all during compilation yet, save for some asserts during codegen. We may want to check these earlier (during unification) sometime in the future for better error messages when users try something like `foo : Ptr` instead of `foo : Ptr i32` for example.

src/types/mod.rs
```diff
@@ -187,6 +188,9 @@ pub const STRING_TYPE: TypeInfoId = TypeInfoId(0);
 /// an operator for its name, but it is otherwise a normal struct type.
 pub const PAIR_TYPE: TypeInfoId = TypeInfoId(1);
 
+/// The Ptr type is like the built in string type
+pub const PTR_TYPE: TypeInfoId = TypeInfoId(2);
```
jfecher, line 192: This should be removed/changed as referenced in a previous comment

jfecher (commented):
Looks good overall, the main change needed is to stick with using the primitive type to refer to Ptr.
And yes, I think pretty much all the extern functions should have `ref` replaced with `Ptr` since C may not always play nicely with region inference. Likewise the `c_string` field of `string` should have its type changed too, I think fixing the above will fix the error message you got

Edit: Regarding tests, you can add some if you want but I won't require them for this PR since Ptrs should already be somewhat indirectly tested via the many tests that use the `c_string` field of `string`. A test specific to pointers would likely use malloc/free and since we already test extern functions and pointer type unification elsewhere I don't see much value in this.

src/nameresolution/builtin.rs
thehabbos007: Ah I see. I must have mixed it up when reading the lines that defined what builtin was.
I spotted in your z3 branch that you did this work for `string` already. Do you want me to exclude it here?

src/nameresolution/builtin.rs
thehabbos007: Thank you! It does make sense to just use the primitive type instead :)

thehabbos007: Since `Ptr` isn't defined as a keyword, how would I specify the type in scope? Can I get around supplying a `TypeInfoId` in 
```rust
    resolver.current_scope().types.insert("Ptr".to_string(), (something that is a TypeInfoId));
```

src/nameresolution/builtin.rs
jfecher: It's probably more appropriate to add it here than my z3 branch. I think I was exploring how easy it'd be to add but forgot to remove it afterward

src/nameresolution/builtin.rs
jfecher: Hmm, yeah I don't think we have a good way of importing that right now. Easiest way for now may be to make `Ptr` a keyword like `string` is

src/nameresolution/builtin.rs
```diff
@@ -70,8 +77,18 @@ pub fn import_prelude<'a>(resolver: &mut NameResolver, cache: &mut ModuleCache<'
...
+    resolver.current_scope().types.insert("string".to_string(), STRING_TYPE);
+    resolver.current_scope().definitions.insert("string".to_string(), BUILTIN_ID);
+
+    // Add pair (,) to scope
     resolver.current_scope().types.insert(Token::Comma.to_string(), PAIR_TYPE);
     resolver.current_scope().definitions.insert(Token::Comma.to_string(), PAIR_ID);
+
+    // Add Ptr to scope
+    resolver.current_scope().types.insert("Ptr".to_string(), PTR_TYPE);
+    resolver.current_scope().definitions.insert("Ptr".to_string(), PTR_ID);
```
jfecher, line 91: One last thing - we shouldn't expose the constructor of `Ptr` here since it says we can create a Ptr by packing any type `a` into a struct which isn't true since the runtime representation of that struct isn't behind a pointer. We can probably avoid defining the `TypeConstructor` for the Ptr type completely.

jfecher (commented):
What is the work remaining after this PR to add in Ptr? Is there any?

thehabbos007:
I don't think there is more after this new push. But it is now ptr as it is a builtin type. The extern functions seems to work fine now as well. Thanks for the help!

src/nameresolution/builtin.rs
thehabbos007: It seems to work just without the whole `define_ptr` function with definition pushing. Should i just remove that alltogether now that `ptr` is a built in type like string? Or does it still need to be in `definition_infos`?

jfecher: Only the constructor needs to be in `definition_infos` and since we don't want one for `Ptr` it's fine to remove `define_ptr` altogether

src/lexer/mod.rs
```diff
@@ -106,6 +106,7 @@ impl<'cache, 'contents> Lexer<'cache, 'contents> {
             ("float", Token::FloatType),
             ("char", Token::CharType),
             ("string", Token::StringType),
+            ("ptr", Token::PointerType),
```
jfecher, line 109: Nitpick: Can we change the keyword to a capital `Ptr` instead? I've been debating whether all types should be capitalized but am still undecided. Since `Ptr` being a keyword may be temporary lets make the keyword capitalized to match its eventual library definition.

src/nameresolution/builtin.rs
```diff
@@ -12,38 +12,49 @@ use crate::types::{ Type, PrimitiveType, TypeInfoBody, Field, LetBindingLevel, S
...
 /// created, otherwise the `builtin` symbol will have the wrong id. If this
 /// happens, this function will assert at runtime.
 pub fn define_builtins<'a>(cache: &mut ModuleCache<'a>) {
-    let string_type = define_string(cache);
-    define_pair(cache);
 
     // Define builtin : forall a. string -> a imported only into the prelude to define
     // builtin operations by name. The specific string arguments are matched on in src/llvm/builtin.rs
     let id = cache.push_definition("builtin", false, Location::builtin());
     assert!(id == BUILTIN_ID);
```
jfecher, line 43: While we're changing this area, can this be a assert_eq?

src/nameresolution/builtin.rs
```diff
@@ -79,30 +90,65 @@ pub fn import_prelude<'a>(resolver: &mut NameResolver, cache: &mut ModuleCache<'
...
+    cache.definition_infos[constructor.0].definition = Some(DefinitionKind::TypeConstructor { name, tag: None });
+
+    cache.type_infos[string_id.0].body = fields;
+    string
 }
 
+/// Defining the 'ptr' builtin type
+///
+/// type ptr a
+fn define_ptr<'a>(cache: &mut ModuleCache<'a>) {
```
jfecher, line 127: Mentioned this in another comment but just so it shows up in the review I'll repeat it here: lets remove this function

jfecher (commented):
Looks good overall, just a few minor things left

src/lexer/mod.rs
thehabbos007: It seemed to try to resolve to a user defined type if it was a capital P. I can try to see if i can change the parsing order for it to resolve correctly though :)

thehabbos007:
That is an odd ci bug. Have you seen that before?

jfecher:
Huh, yeah that is an odd CI bug - haven't seen it before. As long as all the tests pass locally for you I'd say this is ok to merge. Travis-ci.org is closing down on the 15th anyway so I wouldn't bother even investigating this since we'll have to find a new CI solution anyway.

jfecher:
Looks good, thanks for all the revisions!
Hopefully this was an interesting issue since you got to touch most of the different compiler passes in some form.

thehabbos007:
It was great fun to work on, and i feel a little more confident with the code base for the next issues. Thanks for all the help :)

PR #97, opened 2022-05-27
Fix cast char to u8

locriacyber:


jfecher:
This PR removes a `u8: TryFrom<char>` instance ante was using which prevented ante from compilng pre-rustc v1.59! Now it can compile on at least v1.57 as well, thank you!

PR #107, opened 2022-06-22
Fix linker bug where input and output are the same path

0b01:
This PR fixes a bug so cranelift backend works again. Previosuly, the input and output are the same path.

### DIFF
--- a/src/cranelift_backend/context.rs
+++ b/src/cranelift_backend/context.rs
@@ -140,7 +140,7 @@ impl<'local> Context<'local> {
     }
 
     pub fn codegen_all(path: &Path, hir: &'local Ast, args: &Args) {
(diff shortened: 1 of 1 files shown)


jfecher:
Hmm, what was the command where this was failing for you?

0b01:
```bash
root@codespaces-22b8ef:/workspaces/ante# cargo run -- --build examples/codegen/mutability.an 
    Finished dev [unoptimized + debuginfo] target(s) in 0.04s
     Running `target/debug/ante --build examples/codegen/mutability.an`
gcc: fatal error: input file 'examples/codegen/mutability' is the same as output file
compilation terminated.
```

jfecher (approved):

PR #108, opened 2022-06-22
Fix UTF8 strings

ArbilGit:
Currently non-basic UTF8 characters either fail to lex or crash the compiler:

```
s1 = "lumière"
print s1
s2 = "светлый"
print s2
```

This fixes it.


jfecher:
Thank you! Utf8 is a very important consideration I definitely need to test against more in general rather than staying within ascii

PR #112, opened 2022-06-23
Allow local shadowing

0b01:


examples/nameresolution/redeclare.an
```diff
@@ -20,5 +21,5 @@ foo _ =
 // examples/nameresolution/redeclare.an:2:1	note: a previously defined here
 // a = 2
 // 
-// examples/nameresolution/redeclare.an:6:5	warning: a is unused (prefix name with _ to silence this warning)
-//     a = 4
+// examples/nameresolution/redeclare.an:7:5	warning: a is unused (prefix name with _ to silence this warning)
+//     a = 5
```
jfecher, line 25: Hmm, is it possible we can fix this warning? We should still give a warning if a shadowed variable is unused

0b01: The error changed line.

jfecher: I think an easy way to implement it may be to add the unused check right before we overwrite a shadowed variable in scope since it cannot be used after that point.

0b01: Oh I see. Yeah that would be an issue.

jfecher:
Looks good, thank you!

PR #114, opened 2022-06-23
Improve module system

0b01:
This PR refines module system.
1. `import Vec.empty push Vec`
2. `import Vec` imports everything
3. Allows specifying module name on variables and types `Vec.push`

Note the parser is still a little buggy: `import Vec.Vec empty` imports `Vec/Vec.an`.


src/parser/mod.rs
```diff
@@ -73,7 +75,7 @@ fn statement<'a, 'b>(input: Input<'a, 'b>) -> AstResult<'a, 'b> {
     match input[0].0 {
         Token::ParenthesisLeft | Token::Identifier(_) => or(&[definition, assignment, expression], "statement")(input),
         Token::Type => or(&[type_definition, type_alias], "statement")(input),
-        Token::Import => import(input),
+        Token::Import => or(&[import_defs], "import")(input),
```
jfecher, line 78: `or(&[import_defs], ..)(input)` should be equivalent to just `import_defs(input)` with a slightly different error message.

src/parser/mod.rs
```diff
@@ -230,10 +232,11 @@ parser!(struct_inline_body _loc -> 'b ast::TypeDefinitionBody<'b> =
     TypeDefinitionBody::Struct(fields)
 );
 
-parser!(import loc =
+parser!(import_defs loc =
```
jfecher, line 235: Lets keep the old `import` name, the similarity between `import_defs` and `imported_defs` tripped me up while reading this PR

src/parser/mod.rs
```diff
@@ -230,10 +232,11 @@ parser!(struct_inline_body _loc -> 'b ast::TypeDefinitionBody<'b> =
...
     _ <- expect(Token::Import);
-    path <- delimited(typename, expect(Token::MemberAccess));
-    Ast::import(path, loc)
+    path <- delimited_trailing0(typename, expect(Token::MemberAccess));
+    symbols <- many0(imported_defs);
```
jfecher, line 238: While we're changing this section, lets change path and symbols to infallible. E.g. `path !<- delimited_trailing...` and `symbols !<- many0(...)`. Essentially this prevents back tracking once we parse the Import keyword since afterwards we know the only rule import is used for is this one. Should improve error quality for imports a bit.

src/parser/mod.rs
```diff
@@ -580,7 +583,7 @@ fn member_access<'a, 'b>(input: Input<'a, 'b>) -> AstResult<'a, 'b> {
 fn argument<'a, 'b>(input: Input<'a, 'b>) -> AstResult<'a, 'b> {
     match input[0].0 {
         Token::StringType => variable(input),
-        Token::Identifier(_) => variable(input),
+        Token::TypeName(_) | Token::Identifier(_) => or(&[variable, variant], "argument")(input),
```
jfecher, line 586: What is the reasoning for combining the TypeName/variant and Identifier/variable rules into one case?

Edit: Ah because of the shared path I see. Since paths can only consist of TypeNames I think it may be worth keeping the fast case for single variables `Identifier => variable`, and only combining them in the Typename case: `Typename => or([variable, variant])`. Also may be good to make module_path its own parser to avoid redoing work. Then this case can be some equivalent of `module_path then or(variant, variable)` (pseudocode) where neither of variant or variable internally use module_path anymore.

src/parser/mod.rs
```diff
@@ -628,13 +630,17 @@ fn parenthesized_expression<'a, 'b>(input: Input<'a, 'b>) -> AstResult<'a, 'b> {
 }
 
 parser!(variant loc =
-    name <- typename;
-    Ast::type_constructor(name, loc)
+    module_prefix <- delimited(typename, expect(Token::MemberAccess));
+    {
+        let (name, path) = module_prefix.split_last().unwrap();
+        Ast::type_constructor(path.to_owned(), name.to_string(), loc)
```
jfecher, line 636: Can we prevent the cloning here? Should be possible using `pop` instead of split_last

src/parser/pretty_printer.rs
```diff
@@ -32,10 +32,14 @@ impl<'a> Display for ast::Literal<'a> {
 impl<'a> Display for ast::Variable<'a> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         use ast::VariableKind::*;
+        let mut prefix = self.module_prefix.join(".");
```
jfecher, line 35: Need to edit the pretty printing for ast::Import as well

stdlib/String.an
```diff
@@ -0,0 +1,18 @@
+import Vec
+
+type String =
```
jfecher, line 3: File should be removed

jfecher (commented):
Looks pretty good, just a few points:

examples/nameresolution/conflictingimport.an
```diff
@@ -5,15 +5,4 @@ library_int = 1
...
-// 
-// examples/nameresolution/conflictingimport.an:5:1	note: library_fn was previously defined here
-// library_fn _ = 3
-// 
-// examples/nameresolution/conflictingimport.an:1:1	error: import shadows previous definition of library_int
-// import Library
-// 
-// examples/nameresolution/conflictingimport.an:3:1	note: library_int was previously defined here
-// library_int = 1
+// expected stderr:
```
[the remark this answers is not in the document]
0b01: No. If a definition already exists in the current scope, then it's not imported. They can use fully-qualified version though.

src/nameresolution/mod.rs
```diff
@@ -792,14 +798,40 @@ impl<'c> Resolvable<'c> for ast::Variable<'c> {
...
+                }
+                else {
+                    // resolve module
+                    let relative_path = self.module_prefix.join("/");
+                    let module_id = resolver.current_scope().modules.get(&relative_path);
+                    if module_id.is_none() {
+                        if let Some(module_id) = declare_module(Path::new(&relative_path), cache, self.location) {
+                            resolver.current_scope().modules.insert(relative_path.clone(), module_id);
+                            if let Some(exports) = define_module(module_id, cache, self.location) {
+                                resolver.current_scope().import_impls(exports, cache);
```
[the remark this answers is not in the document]
0b01: Done.

src/nameresolution/mod.rs
```diff
@@ -107,6 +108,9 @@ pub struct NameResolver {
     /// be defined until the 'define' pass later however.
     pub exports: Scope,
 
+    /// contains scope of imported modules, contains definitions, impls, etc.
+    pub module_scopes: HashMap<ModuleId, Scope>,
```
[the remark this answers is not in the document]
0b01: The module scope is stored here to lookup DefinitionId

src/nameresolution/scope.rs
```diff
@@ -70,32 +75,24 @@ impl Scope {
...
-    fn import_definitions_types_and_traits(&mut self, other: &Scope, cache: &mut ModuleCache, location: Location) {
+    fn import_definitions_types_and_traits(&mut self, other: &Scope, cache: &mut ModuleCache, location: Location, symbols: &HashSet<String>) {
         macro_rules! merge_table {
-            ( $field:tt , $cache_field:tt , $errors:tt ) => {{
+            ( $field:tt , $cache_field:tt , $errors:tt, $symbols:expr ) => {{
                 for (k, v) in other.$field.iter() {
-                    if let Some(existing) = self.$field.get(k) {
-                        let prev_loc = cache.$cache_field[existing.0].locate();
-                        let error = make_error!(location, "import shadows previous definition of {}", k);
-                        let note = make_note!(prev_loc, "{} was previously defined here", k);
```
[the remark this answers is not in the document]
0b01: Global shadowing is still not allowed with the exception of importing modules. If a definition already exists, skip silently.

src/parser/combinators.rs
```diff
@@ -28,6 +28,12 @@ macro_rules! seq {
         let ($input, $name, start) = $y($input)?;
         seq!($input start start $location => $($rest)*)
     });
+    // This rule specifically is the first bind that occurs, hence why
+    // the location here is forwarded to the rest of the macro as the start location.
```
jfecher, line 32: Lets update this comment, we can just say it is a mutable version of the previous rule

src/util/mod.rs
```diff
@@ -39,7 +39,7 @@ pub fn unwrap_clone<T: Clone>(option: &Option<T>) -> T {
 }
 
 /// Convert each element to a String and join them with the given delimiter
-pub fn join_with<T: Display>(vec: &[T], delimiter: &str) -> String {
+pub fn join_with<T: Display>(vec: impl IntoIterator<Item = T> , delimiter: &str) -> String {
     fmap(vec, |t| format!("{}", t)).join(delimiter)
```
jfecher, line 43: Lets rename `vec` to `items` or similar to reflect it is no longer required to be an array

stdlib/prelude.an
```diff
@@ -263,8 +263,8 @@ impl TryCast u64 i64 with
         if x > 9_223_372_036_854_775_807 then None
         else Some (transmute x)
 
-trait Append t with
-    (++): t - t -> t
+trait Append a b with
+    (++): a - b -> a
```
jfecher, line 267: What is the reasoning on this change?
Requiring both operands to be the same type makes it a Semigroup and was a deliberate decision to aid type inference, which makes impls easier to find. I think we should keep `Append t` until we have a good usecase for a mixed append at least. Potentially `Append StringBuilder string` in the future, but we may also want to generalize the return type as well.

jfecher (commented):
Looks very good, and is a large improvement to the current system! Thanks for taking the time to implement all this.

There is still some further work in the future, e.g. reading this PR I believe it is now the case that using a fully-qualified variable `Foo.Bar.baz 3` will also import all the impls from `Bar` into local scope. If I'm correct on this, we can look into fixing it later since the other features enabled by this PR are more important to me. The time to revisit it may be the same time when named impls are implemented so they can be handled the same way as other imported items - by name.

A few remaining comments, most are minor:

jfecher:
Thank you! 👍

PR #116, opened 2022-06-29
Add build.rs to set ANTE_STDLIB_DIR

mtshiba:
I ran the compiler somewhere other than the project root, and it crashed.

```console
$ ante test.an
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 3, kind: NotFound, message: "No such file or directory" }', src\util\mod.rs:82:33
stack backtrace:
...
```

The cause of it is `stdlib_dir()` in util/mod.rs.

```rust
pub fn stdlib_dir() -> PathBuf {
    match option_env!("ANTE_STDLIB_DIR") {
        Some(env) => std::fs::canonicalize(env).unwrap(),
        None => {
            let mut path = PathBuf::from(file!());
            path.pop();
            path.pop();
            path.pop();
            path.push("stdlib");
            path.canonicalize().unwrap()
        }
    }
}
```

This doesn't work if `ANTE_STDLIB_DIR` is not set and the compiler is executed other than the project root, since `file!()` doesn't return the absolute path.

So I created a build script that copies `stdlib` and sets `ANTE_STDLIB_DIR`.

For now, `stdlib` is located at `$HOME/.ante`. If you don't like this, tell me the appropriate location.

For reference, Rust's standard library is located at `.rustup/toolchains/<arch>/lib/rustlib`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "ante"
 version = "0.1.1"
 authors = ["Jake Fecher <jfecher11@gmail.com>"]
 edition = "2018"
+build = "build.rs"
 
 [lib]
 name = "ante"
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,42 @@
+#![allow(deprecated)]
+
+use std::env;
+use std::fs;
+use std::path;
+
+fn main() -> std::io::Result<()> {
+    // env::home_dir() is deprecated, maybe we should use dirs::home_dir()
+    let dot_ante = env::home_dir()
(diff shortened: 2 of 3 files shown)


jfecher:
This used to be the default before `build.rs` was changed recently. Instead of having a build.rs default to a home directory, I think we should fix `stdlib_dir` - does using [std::path::absolute](https://doc.rust-lang.org/std/path/fn.absolute.html) work?

mtshiba:
OK, I fixed.

jfecher:
👍

PR #117, opened 2022-06-30
Do nothing if the input file is empty

mtshiba:
This is a trivial problem, but the compiler crashed when I tried to compile an empty file.

```console
$ ante empty.an
empty.an:1:1     error: failed trying to parse a term
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src\error\mod.rs:203:79
stack backtrace:
...
```

In languages that do not have or require entry points, such as the main function, an empty file is often valid input and nothing is done as a result of execution.

So I've modified the compiler to immediately terminate if the input is an empty file.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "ante"
 version = "0.1.1"
 authors = ["Jake Fecher <jfecher11@gmail.com>"]
 edition = "2018"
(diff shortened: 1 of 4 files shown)


mtshiba:
I made a mistake. I committed before the previous PR was merged.

I will close this PR once.

PR #118, opened 2022-06-30
Do nothing if the input file is empty

mtshiba:
This is a trivial problem, but the compiler crashed when I tried to compile an empty file.

```console
$ ante empty.an
empty.an:1:1     error: failed trying to parse a term
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src\error\mod.rs:203:79
stack backtrace:
...
```

In languages that do not have or require entry points, such as the main function, an empty file is often valid input and nothing is done as a result of execution.

So I've modified the compiler to immediately terminate if the input is an empty file.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -92,6 +92,9 @@ pub fn main() {
     let filename = Path::new(&args.file);
     let file = File::open(filename);
     let file = expect!(file, "Could not open file {}\n", filename.display());
+    if let Ok(0) = file.metadata().map(|data| data.len()) {
+        return;
+    }
 
     let mut cache = ModuleCache::new(filename.parent().unwrap());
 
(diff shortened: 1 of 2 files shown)


mtshiba:
I noticed that this is not a complete solution, it still crashes if the imported file is empty.

mtshiba:
Fixed to ignore imported files if they are empty.

jfecher:
I don't think a hard coded check for if the file length is 0 is the right way to solve this problem. The crash occurs on an `unwrap` while reporting an error, so I pushed a commit changing it to `unwrap_or("")` which solves it.

There remains the separate issue of running the compiler on an empty file (or one only filled with whitespace or comments) producing a parser error:
```
empty.an:1:1	error: failed trying to parse a term
```
I consider this a separate issue, so I'm closing this PR as the core issue is fixed.

Issue #82, opened 2021-05-22
New looping sugar


jfecher:
Since I'm still undecided on whether ante should have dedicated loop constructs like `while` and `for` or not I've been looking into implementing some looping functions instead. Replacing `for` is fairly simple, but `while` can be more difficult in part because these loops somewhat rely on mutation. Another important aspect is these looping constructs also work with `break` and `continue` which most helper functions don't.
I've looked to other languages - particularly haskell and purescript for ideas and am thinking about adding some functions into the prelude based on purescript's MonadRec module: https://pursuit.purescript.org/packages/purescript-tailrec/4.1.1/docs/Control.Monad.Rec.Class#t:MonadRec

Currently, I'm only planning on adding a few core functions. Here are some examples in ante:
```scala
type Loop c b =
   | Continue c
   | Break b

loop (initial: c) (f: c -> Loop c b) : b =
    match f initial with
    | Continue c -> loop c f
    | Break b -> b
    
loop2 (x: x) (y: y) (f: x -> y -> Loop (x, y) b) : b =
    loop (x, y) fn (x, y) -> f x y

// Simple usecase
sum (numbers: List Int) : Int =
    loop2 0 numbers fn total remaining ->
        match remaining with
        | Nil -> Break total
        | Cons x xs -> Continue (total + x, xs)

// Breaking early
given Iterator iter elem
index_of (iterator: iter) (element: elem) : Maybe usz =
    loop2 iterator 0 fn iterator index ->
        match next iterator with
        | None -> Break None
        | Some (x, rest) if x == element -> Break (Some index)
        | Some (_, rest) -> Continue (rest, index + 1)
```
These functions are more general than iterator functions and could thus be used to implement iterators (as seen in the implementation of `index_of` above).

Further questions on this proposal include:
- Is there a more general looping function than `loop` we should build on instead?
- `loop` is a very common name, is the use of this function common enough for it to use that name?
- Should there be a `loop3` as well?
- Does imperative `while`-looping code translate well into `loop` or is it too cumbersome?

jfecher:
Another option over `loop` with a `Loop` data type is to instead pass another function as a recursive callback:
```scala
tail_recur (initial: c) (f: c (c -> r) -> r) : r =
    f2 x = f x f2
    f2 initial
    
tail_recur2 (first: x) (second: y) (f: x y (x y -> r) -> r) : r =
    f2 x y = f x y f2
    f2 first second

sum (numbers: List Int) : Int =
[...]

PR #120, opened 2022-06-30
Loop syntax

ArbilGit:
Implements loop desugaring syntax. 

Addresses #82


jfecher (commented):
Looks pretty good! And yes, regarding your comment I believe the grammar for loop arguments should be an identifier or parenthesized(pattern = expr)

jfecher:
Type annotations are part of pattern syntax, so those should work without additional cases

jfecher:
There's been no development on this for a little while and the remaining changes were small so I went ahead and opened #121 which implements the remainder of this PR.

Issue #125, opened 2022-09-13
Include a nix flake in the root of the repository


ehllie:
Including a `flake.nix` with appropriate expressions inside it, would allow people that use the nix package manager to install ante directly from this repository. I've made a PR to include ante in the main nixpkgs repository already, but merging that pull request seems to be taking some time. It would be convenient to offer an alternative with the bleeding edge version of the compiler to people that prefer to use that over the version available in the official repositories.

A flake would also allow for easily reproducible developer environments, and make non-rust dependencies this project has like `llvm` much easier to configure properly, needing only to run `nix build` to have the package manager install all dependencies and build the compiler, or `nix develop` to enter an already configured development environment where `cargo build` will correctly run.

ehllie:
I suppose I ought to add a section in the readme mentioning you can build the project with nix as well, now that the PR was merged

PR #126, opened 2022-09-13
Add nix flake and package derivation

ehllie:
#125
This provides a flake with `overlays` and `packages` for each default platform in its outputs, as well as a `default.nix` with the ante package derivation which should work with non nix-command cli.


jfecher (approved):
Thank you!

(apologies for the delay)

Issue #83, opened 2021-05-22
`impl Cast int string` in the stdlib should compile


jfecher:
Currently, `impl Cast int string with ...` is one of the commented out sections of the prelude since it hasn't worked since the start of the rewrite of the old compiler from C++ into Rust. I think now is the time to start working on adding any small features needed to get more of these commented out sections to compile - starting with this one. Here is the current version of `Cast int string`:
```scala
impl Cast int string
    cast i =
        if i == 0 then return "0"
        len = mut 0
        alloc_size = 20
        buf = mut malloc (alloc_size + 1)
        buf#alloc_size := '\0'
 
        x = mut i
        isNeg = i < 0 and ((x *= -1); true)

        while x != 0 do
            len += 1
            buf#(alloc_size - len) := '0' + cast (x % 10)
            x /= 10

        if isNeg then
            len += 1
            buf#(alloc_size - len) := '-'

        offset = cast buf + (alloc_size - len)
        string (cast offset) len
```
For it to compile, several changes will need to be made:
- [x] New impl syntax, should be `impl Cast int string with ...`
- [x] We should implement it for all integers now that we have many: `impl Cast int string given Int int with ...`
- [ ] Ante doesn't have loops anymore, so the `while` loop will need to be replaced with a recursive helper function. ~~Ante also doesn't have closures yet so this helper may be more verbose than otherwise necessary.~~
- [x] The final `string (cast offset) len` call to manually construct a string isn't yet possible since the parser doesn't accept `string` as a valid value yet - this should probably be fixed so it parses correctly.
- [x] The final and most difficult requirement on this list is to get pointers and indexing (via `#`) working. A minimum viable product for this can use a normal named function over `#`, though a method to offset pointers or `ref`s will still be necessary since this code calls `malloc` manually.
    - [x] An `offset` function for pointers will likely require adding a compiler builtin in `llvm/builtin.rs`

thehabbos007:
Would it simplify the implementation of this to sort out #82 first? :)

Also would you prefer this or the Ptr primitive type to be taken care of first?

jfecher:
#82 is mostly unrelated - using `loop` over a recursive function for the while loop would be a bit simpler but not enough to justify waiting on #82 imo.
[...]

PR #127, opened 2022-09-22
Implement generic int to string casts

ehllie:
Implements casts to string for generic int types in the stdlib, as well as adds a couple simple tests for them #83

### DIFF
--- a/examples/codegen/int_string_cast.an
+++ b/examples/codegen/int_string_cast.an
@@ -1,0 +1,18 @@
+debug str =
+  print <|
+    "string: " ++
+    str ++
+    " len: " ++
+    cast str.length
+
+debug <| cast (255: u8)
+
+debug <| cast -1024
+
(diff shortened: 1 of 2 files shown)


examples/codegen/int_string_cast.an
```diff
@@ -0,0 +1,31 @@
+max_u8: u8 = 255
+
+smu: string = cast max_u8
+
+print <|
+  "string: " ++
+  smu ++
+  " len: " ++
+  cast smu.length
```
jfecher, line 9: Lets put this printing in a helper function to avoid repeating it. I think it'd help to put the focus more on the int to string cast rather than the debug printing which currently takes up most of the file

jfecher (commented):
Looks good, I appreciate calculating the length of the resulting string via repeated multiplication rather than division

jfecher (approved):
👍

PR #128, opened 2022-09-22
Update readme with a section on Nix

### LINKED ISSUE #125: Include a nix flake in the root of the repository

ehllie:
Considering #126 was merged, it wouldn't hurt to mention the possibility of using nix for building, as it should make it trivial for anyone to setup the environment and or install ante.
Closes #125

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ than development updates. You can also feel to file issues or ask questions on t
 
 Ante currently optionally requires llvm 13.0 while building. If you already have this installed with
 sources, you may be fine building with `cargo install --path .` alone. If cargo complains
-about not finding any suitable llvm version, you can either choose to build ante without 
+about not finding any suitable llvm version, you can either choose to build ante without
 the llvm backend via `cargo install --path . --no-default-features` or you can build llvm from
 source, either via `llvmenv` or `cmake` as covered in the next sections.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,6 +102,20 @@ $ cargo build
 You can confirm your current version of llvm by running `llvmenv version`
 or `llvm-config`
 
+#### Nix
+
+Ante is available in the unstable branch of the [nixpkgs repository](https://search.nixos.org/packages?channel=unstable&show=ante&type=packages&query=ante).
```
jfecher, line 107: I think this should be a subsection of the linux section rather than being after the windows section

ehllie: I suppose that would make more sense

jfecher (approved):
👍

Issue #91, opened 2022-03-26
Implement string interpolation


jfecher:
String interpolation is the last feature from the old C++ compiler that is still unimplemented in the rust compiler.

Once implemented this feature should desugar strings in the form of `"... ${expr} ..."` into `"... " ++ cast expr : string ++ " ..."`. This looks like the following in the Ast:
```rs
FunctionCall(Variable("++"), vec![
    StringLiteral("... "),
    FunctionCall(Variable("++"), vec![
        TypeAnnotation(
            FunctionCall(Variable("cast"), vec![expr]),
            Type::String,
        ),
        StringLiteral(" ..."),
    ])
])
```

The first and last parts of the string can be elided if they're empty. We may also want to desugar strings in the form `"... $identifier ..."` to work equivalently to `"... ${identifier} ...`.

An implementation of this feature should likely perform this desugaring when constructing the string node while parsing - checking if the string contains `$` and so on. We'll also need to allow escaping `$` via `\`. As a design note, it may be that the `" $identifier "` syntax collides too much with users expecting to use `$` characters in the string. If this is the case only allowing the `${expr}` syntax may be preferable.

It may be in the future we want a more advanced desugaring so that we can e.g. desugar to some `LazyInterpolate` type to avoid constructing strings that will be immediately logged. This is an optimization that can wait.

ehllie:
Should all strings be implicitly formatted like this, or should we do it like python does it for example, and only interpolate strings prefixed with `f` or using some other identifier?
```python
foo = "Hello"
"{foo}, world!" # "{foo}, world!"
f"{foo}, world!" # "Hello, world!"
```

ehllie:
Here's a couple examples of how other languages have done this:

### Python
Like i mentioned above, python uses fstring format, where you can interpolate any expression inside a string prefixed with `f` by wrapping it in `{expr}`
```python
foo = lambda x: x + ", World"
bar = "Hello"
print(f"{foo(bar)}!")  # Hello, World!
```

### JavaScript
JavaScript uses backticks, and then indicates expressions to interpolate with a `${expr}` syntax.
```js
foo = (x) => x + ', World'
bar = 'Hello'
console.log(`${foo(bar)}!`) // Hello, World!
```

### Rust
Rust uses macros to achieve string interpolation, and the expressions to interpolate are passed in as arguments to the macro
```rust
fn main() {
    fn foo(x: &str) -> String {
        let mut ret = String::from(x);
        ret.push_str(", World");
        ret
[...]

PR #131, opened 2022-09-25
Implemented string interpolation

ehllie:
#91
- Fixed a bug with int to string casting where `10`, `100`, etc would get cast to `"0"`, `"00"`, etc.
- Fixed a bug where the lexer would continue beyond EOF if the closing quote was not found.
- Fixed a bug where the lexer escaped `\'` instead of `\"` which are used for string literals.
- Added two new lexer tokens for indicating a start and end of an interpolation.
- Strings are parsed as `StringLiteral` and an any number of `InterpolateLeft` `expression` `InterpolateRight` and `StringLiteral` pairs.
- Expressions can be interpolated into strings by wrapping them with `${}`.

---

I also want to drop empty string literals when doing interpolations since currently any interpolated expression is followed by a string literal, meaning `"${a}${b}${c}"` gets parsed to `('++' ('++' ('++' "" ('++' (cast a) "")) ('++' (cast b) "")) ('++' (cast c) ""))`

### DIFF
--- a/examples/parsing/string_interpolation.an
+++ b/examples/parsing/string_interpolation.an
@@ -1,0 +1,19 @@
+shout x = "${x}!"
+
+bar = "nested interpolations"
+
+print
+  "I can do line breaks${
+    shout "\nAny expressions"
+  }${
+    shout "\nEven ${bar}"
+  }"
+
+// args: --parse
+// expected stdout:
+// (shout = (fn x -> ('++' "" ('++' (cast x) "!"))));
+// (bar = "nested interpolations");
+// (print ('++' ('++' "I can do line breaks" (cast (shout "
+// Any expressions"))) (cast (shout ('++' "
+// Even " (cast bar))))))
+
--- a/examples/parsing/string_interpolation.an
+++ b/examples/parsing/string_interpolation.an
@@ -11,7 +11,7 @@ print
 
 // args: --parse
 // expected stdout:
-// (shout = (fn x -> ('++' "" ('++' (cast x) "!"))));
+// (shout = (fn x -> ('++' (cast x) "!")));
 // (bar = "nested interpolations");
 // (print ('++' ('++' "I can do line breaks" (cast (shout "
 // Any expressions"))) (cast (shout ('++' "
--- a/examples/parsing/string_interpolation.an
+++ b/examples/parsing/string_interpolation.an
@@ -11,7 +11,7 @@ print
 
 // args: --parse
 // expected stdout:
-// (shout = (fn x -> ('++' (cast x) "!")));
+// (shout = (fn x -> (: ('++' (cast x) "!") string)));
 // (bar = "nested interpolations");
 // (print ('++' ('++' "I can do line breaks" (cast (shout "
 // Any expressions"))) (cast (shout ('++' "
--- a/src/lexer/mod.rs
+++ b/src/lexer/mod.rs
@@ -45,6 +45,13 @@ use std::str::Chars;
 use token::{IntegerKind, LexerError, Token};
 
 #[derive(Clone)]
+struct OpenBraces {
+    parenthesis: usize,
+    curly: usize,
+    square: usize,
+}
+
+#[derive(Clone)]
 pub struct Lexer<'cache, 'contents> {
     current: char,
     next: char,
@@ -58,6 +65,8 @@ pub struct Lexer<'cache, 'contents> {
     previous_token_expects_indent: bool,
     chars: Chars<'contents>,
     keywords: HashMap<&'static str, Token>,
+    open_braces: OpenBraces,
+    pending_interpolations: Vec<usize>,
 }
 
 /// The lexer maintains a stack of IndentLevels to remember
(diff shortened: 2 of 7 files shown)


ehllie:
I'm making the PR early, before I've done everything yet, cause I'd like to know whether the rest of the code is ok, or if I went in the completely wrong direction with it

ehllie:
The AST should be a little cleaner with empty strings pruned when interpolating now.

examples/parsing/string_interpolation.an
```diff
@@ -0,0 +1,19 @@
...
+print
+  "I can do line breaks${
+    shout "\nAny expressions"
+  }${
+    shout "\nEven ${bar}"
+  }"
+
+// args: --parse
+// expected stdout:
+// (shout = (fn x -> ('++' "" ('++' (cast x) "!"))));
```
jfecher, line 14: I'd definitely like to see this simplified to
```rs
(shout = (fn x -> ('++' (cast x) "!")));
```
Without the extra empty string literal if possible. Seems like a check could be added in the desugar parser functions for it. 

Also, although it should be inferred the majority of the time, I believe the `: string` may still be necessary to add in the desugaring after casting. My thinking is given a literal such as `"${foo}"` if we simplify out the empty string literal appendings we are left with `cast foo` as the desugared code which may not always resolve to a string without the extra annotation.

src/lexer/mod.rs
```diff
@@ -323,22 +323,26 @@ impl<'cache, 'contents> Lexer<'cache, 'contents> {
     fn lex_string(&mut self) -> IterElem<'cache> {
         self.advance();
         let mut contents = String::new();
-        while self.current != '"' {
-            let current_char = if self.current == '\\' {
-                self.advance();
-                match self.current {
-                    '\\' | '\'' => self.current,
```
jfecher, line 330: Good catch

src/lexer/mod.rs
```diff
@@ -500,6 +504,11 @@ impl<'cache, 'contents> Iterator for Lexer<'cache, 'contents> {
                 }
             },
             ('"', _) => self.lex_string(),
+            ('}', _) => {
+                self.current = '"';
+                Some((Token::InterpolateRight, self.locate()))
+            },
+            ('$', '{') => self.advance2_with(Token::InterpolateLeft),
```
jfecher, line 511: This was a rather clever implementation I didn't think of originally. It does mean that we rely on the interpolation characters not being valid in the regular expression language however. This is a relatively minor problem, though it does limit the ending character `}` from being used - we couldn't implement the `"\(foo)"` syntax with this for example.

I was wondering if this could cause worse error messages for users that try to use `${` or `}` outside of strings. The most impact it'd have I think should be mentioning an "unexpected '${' or '}' found, expected ...." which seems fine. This method doesn't work for nested interpolation like `"${foo "${bar}" baz }"` but I think that is fine.

jfecher (commented):
Looks pretty good and handles nested string literals inside interpolated strings well. My one worry is that accepting `${` and `}` as separate tokens in the lexer in all contexts means we cannot change the syntax in the future to something like `\(` and `)` since the `)` would have to be valid for other contexts besides an InterpolateRight token as well.

I think one way around this is to keep the string literal as one big literal (possibly with some special handling for nested quotes? If not they can still be escaped normally) and re-lex/parse within the desugar method. This may have its own problems though and the current solution does work for the current syntax so I'll leave the decision to you.

src/lexer/mod.rs
ehllie: One approach i thought of was to keep track of the parenthesis level with the lexer, and only issue the InterpolationRight token only when the parenthesis level is zero. That would allow us to switch to a different interpolation syntax and use the token `}`outside of string interpolations

jfecher: That's a good solution - we should use it for InterpolateLeft as well

examples/parsing/string_interpolation.an
ehllie: Empty leading strings are now pruned in the concatenation, and if that happens a type annotation gets added.

src/lexer/mod.rs
ehllie: We should be able to now still use the `}` token outside of string interpolation if we ever want to, or switch to `\()` syntax for interpolation. The lexer will issue `InterpolateRight` if the number of brackets responsible for opening the bracket to which `InterpolateRight` is mapped to that have not been closed, is the same as when `InterpolateLeft` token was last issued.

ehllie: Actually, separating open brackets into subtypes might be unnecessary. We could just keep a stack of all open brace tokens, and check if it's length matches. That might also also allow for better error messages since we could check for braces matching when popping the open braces. Not sure whether that would go beyond the scope of a lexer however

jfecher: Hmm, yeah looking over the most recent commit I do think it may be overkill to track all matching pairs. Is it really necessary to track any other tokens other than `${` and `}` for this? I was thinking the lexer could have a single `interpolate_level: u32` field or similar.

jfecher (approved):
There has been some good exploration done on alternatives, and although there may be some possible simplifications to the lexer, these can be done in a later PR. I think it is time to stop delaying this one and merge it since it implements a good feature. Thank you!

PR #132, opened 2022-09-27
Interpolating strings by re-running lexer and parser

ehllie:
#91
Alternative to #131.
This does not introduce any extra tokens to the lexer, and should keep the lexer logic simpler by moving the interpolation logic inside closures, and should hopefully be quite future proof. 

The downsides to this approach is doing extra computation, since each string gets read at least twice, and as long as error messages rely on explicit file line and column numbers being passed around, syntax errors inside interpolations might be not the most helpful.

The former should not be a massive problem, especially if we introduce a syntax for non-interpolated-non-escaped at some later point, and the latter can be fixed.

### DIFF
--- a/examples/codegen/int_string_cast.an
+++ b/examples/codegen/int_string_cast.an
@@ -11,8 +11,11 @@ debug <| cast -1024
 
 debug <| cast 0
 
+debug <| cast 10
+
 // args: --delete-binary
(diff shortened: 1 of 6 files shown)


jfecher:
Hmm, I think I do prefer the other method after all. Thanks for exploring this!

Issue #135, opened 2022-10-04
Improve CLI


ehllie:
Ante CLI is quite basic, and allows for use of contradictory options, like using both `--check` and `--backend <BACKEND>` at the same time, even though the latter will have no effect on what the program does. There's also no way to generate syntax completion for it.

jfecher:
Closing this after the merge of #136

PR #136, opened 2022-10-04
CLI improvements

ehllie:
### Ideas for changes so far

Overhauling the cli by splitting the functionality into 3 sub-commands:

- `check`: Equivalent of `ante --check` and has options for printing out the result of each step before codegen.
- `build`:  Equivalent of `ante --build` and has the options for configuring the backend and optimisation level.
- `run`: The current default `ante`, has the same options as `build`, but also one to delete the binary afterwards.

Additionally, a new functionality inside the sub-command `completions` which just allows for automatically generating code completion scripts with `clap_complete`

#135


ehllie:
https://user-images.githubusercontent.com/20847625/194134698-357cc935-2188-4018-8a27-ccb184a45494.mp4

This is what I have so far

ehllie:
I can alternatively try doing something like this:

- Group `build`, `delete-binary`, `show-hir`, `show-ir`, `backend` and `opt_level` flags and options into a codegen args struct
- Make `build` and `delete-binary` flags mutually exclusive in that struct
- Make `check` flag mutually exclusive with the codegen args fields
- Make a `print-completion` option that's mutually exclusive with all other options

That should preserve the current functionality, add some extra check and allow for generating syntax completion.

jfecher:
That sounds better. As far as prioritization goes, I think the most important aspects are (in order):
1. Getting some command line completion working
2. Maintaining the current command line syntax, or keeping it flexible for the future. E.g. a large change that will eventually be necessary is forwarding all arguments after the filename when running a file to the argv of that program. So that you could invoke it via `ante myecho.an this is echoed`.
3. Ease of using the clap/cli API for the developer, brevity of code etc. If it requires too much code it may not be worth it for such a simple thing.
4. Least important to me is preventing mutually conflicting options like `check` and `run`. I think most of these should be fairly obvious to users that they are conflicting, but an explicit error would always be nice.

jfecher (approved):
Looks good, thank you!

Issue #92, opened 2022-03-27
Make `if` without `else` return `Maybe a`


jfecher:
Currently, if without an else always throws away its value and returns `()`.

This is unnecessary - having it instead return `Some value` if the then branch was taken fits more inline with an expression-oriented language and obviates the need for functions like `bool::then` in rust.

jfecher:
Based on experimentation in #143 and the `loop-sugar` branch it seems this would create issues with using the pattern `loop () -> if ... then recur ()` since the inferred type of the loop would be an infinite type `a = Maybe a`. Closing this PR for now.

PR #143, opened 2022-11-03
Draft: Issue #92 Allow `Maybe` type to be retuned to `If` if no `else` is defined

anoojpatel:
This PR attempts to add `MAYBE` type into the `builtins.rs` for `Ast::If`
- [ ] Tests

### DIFF
--- a/src/nameresolution/builtin.rs
+++ b/src/nameresolution/builtin.rs
@@ -9,7 +9,7 @@ use crate::error::location::Location;
 use crate::lexer::token::{IntegerKind, Token};
 use crate::nameresolution::{declare_module, define_module, NameResolver};
 use crate::types::{
-    Field, FunctionType, GeneralizedType, LetBindingLevel, PrimitiveType, Type, TypeInfoBody, MAYBE_TYPE, PAIR_TYPE, STRING_TYPE,
+    Field, FunctionType, GeneralizedType, LetBindingLevel, PrimitiveType, Type, TypeConstructor, TypeInfoBody, MAYBE_TYPE, PAIR_TYPE, STRING_TYPE,
 };
 
 use std::collections::HashSet;
@@ -28,7 +28,7 @@ pub const STRING_ID: DefinitionInfoId = DefinitionInfoId(1);
 pub const PAIR_ID: DefinitionInfoId = DefinitionInfoId(2);
 
 /// DefinitionInfoId for the Maybe to construct values like Maybe a
-pub const MAYBE_ID: DefinitionInfoId = DefinitionInfoId(2);
+pub const MAYBE_ID: DefinitionInfoId = DefinitionInfoId(3);
 
 /// Defines the builtin symbols:
 /// - `type string = c_string: ptr char, len: usz`
@@ -200,8 +200,8 @@ fn define_maybe(cache: &mut ModuleCache) {
     let maybe = cache.push_type_info(name.clone(), vec![], location);
     assert_eq!(maybe, MAYBE_TYPE);
 
-    cache.type_infos[maybe.0].body = TypeInfoBody::Struct(vec![
-        Field { name: "value".into(), field_type: Type::TypeVariable(a), location },
+    cache.type_infos[maybe.0].body = TypeInfoBody::Union(vec![
+        TypeConstructor { name: "value".into(), args: vec![Type::TypeVariable(a), Type::UNIT], id: MAYBE_ID, location },
     ]);
 
     cache.type_infos[maybe.0].args = vec![a];
@@ -226,7 +226,7 @@ fn define_maybe(cache: &mut ModuleCache) {
     // and now register a new type constructor in the cache with the given type
     let id = cache.push_definition(&name, location);
     assert_eq!(id, MAYBE_ID);
-    let constructor = DefinitionKind::TypeConstructor { name, tag: None };
+    let constructor = DefinitionKind::TypeConstructor { name, tag: Some(0) };
 
     cache.definition_infos[id.0].typ = Some(constructor_type);
     cache.definition_infos[id.0].definition = Some(constructor);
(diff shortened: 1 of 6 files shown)


src/nameresolution/builtin.rs
```diff
@@ -179,3 +184,51 @@ fn define_pair(cache: &mut ModuleCache) {
...
+
+    let level = LetBindingLevel(0);
+    let a = cache.next_type_variable_id(level);
+
+    let name = "Maybe".to_string();
+    let maybe = cache.push_type_info(name.clone(), vec![], location);
+    assert_eq!(maybe, MAYBE_TYPE);
+
+    cache.type_infos[maybe.0].body = TypeInfoBody::Struct(vec![
+        Field { name: "value".into(), field_type: Type::TypeVariable(a), location },
```
jfecher, line 204: Maybe should be a union type, not a struct type. Since its value can be either Some a or None

src/types/typechecker.rs
```diff
@@ -1734,7 +1734,7 @@ impl<'a> Inferable<'a> for ast::If<'a> {
             unify(&then.typ, &otherwise.typ, self.location, cache, msg);
             result.with_type(then.typ)
         } else {
-            result.with_type(Type::UNIT)
+            result.with_type(Type::UserDefined(MAYBE_TYPE))
```
jfecher, line 1737: Looks like you're returning a Maybe type constructor here without applying it to the relevant type arguments. E.g. it is like you're returning `Ptr` instead of `Ptr Int`. You'll likely want to return a TypeApplication of Maybe and then.typ.

We also may need codegen support for this as well....
Perhaps an easier approach for now may be to desugar `if a then b` into `if a then Some b else None` instead.

anoojpatel:
I got some time to take a crack at the parser desugaring and got [this so far. ](https://github.com/jfecher/ante/pull/143/files#diff-b78363c945458ba512de20a9df6c5b2536b373fd115e181701f89ee0e7acf6d3R554-R564 ) . I'm seeming to run into codegen issues 
```
➜ ante test.an
thread 'main' panicked at '- inst5 (jump block3(v3)): arg 0 (v3) has type f64, expected i8
- inst5 (jump block3(v3)): mismatched argument count for `jump block3(v3)`: got 1, expected 3
- inst7 (jump block3(v7)): mismatched argument count for `jump block3(v7)`: got 1, expected 3
', src/cranelift_backend/context.rs:194:13
[1]    53622 abort      ante test.an
```
Should I be trying to infer the type differently? My approach is to desugar into unit and then on the typing side, infer the then, and check if otherwise returns unit (in the case we have no else) to then return the custom Maybe. does approach make sense?

src/parser/ast.rs
```diff
@@ -549,13 +549,19 @@ impl<'a> Ast<'a> {
...
+        match &otherwise{
+            Some(_ast) => Ast::If(If {
             condition: Box::new(condition),
             then: Box::new(then),
             otherwise: otherwise.map(Box::new),
             location,
             typ: None,
-        })
+        }),
+            None => Ast::if_expr(condition,then, Some(Ast::unit_literal(location)), location)
```
jfecher, line 562: If you're trying to do the desugaring, you're going about this a bit differently than you should. Going this route for example you should only need to change this function - you don't need to touch type checking at all.

There's also an issue with what you're desugaring too currently. Currently in the None case you're desugaring `if a then b` into `if a then b else ()` while you should be desugaring to `if a then Some b else None` instead.

src/parser/ast.rs
```diff
@@ -549,13 +549,19 @@ impl<'a> Ast<'a> {
...
 
     pub fn if_expr(condition: Ast<'a>, then: Ast<'a>, otherwise: Option<Ast<'a>>, location: Location<'a>) -> Ast<'a> {
-        Ast::If(If {
+        
+ 
+        match &otherwise{
+            Some(_ast) => Ast::If(If {
             condition: Box::new(condition),
             then: Box::new(then),
             otherwise: otherwise.map(Box::new),
```
jfecher, line 558: If you're explicitly matching on `otherwise` here now, then you can use the `_ast` directly instead of ignoring it and doing `otherwise.map(...)` later on.

jfecher (commented):
Still some work required conceptually but the good news is you can remove 90% of your changes and only focus on the `if_expr` function

src/parser/ast.rs
anoojpatel: Gotcha! Again thanks for the guidance and patience 😅 ! 
> Currently in the None case you're desugaring if a then b into if a then b else () while you should be desugaring to if a then Some b else None instead.
Would this mean I need to modify the `struct If` in parser/ast.rs such that `pub then: Option<Box<Ast<'a>>>` to allow `then` to be Optional (given in your response, we'd want to Desugar `Some b`)?

jfecher: No - I think you're having some confusion here between the language the compiler is written in - rust - and the language we want to compile - Ante. From the compiler's perspective the `If` struct should represent an if expression in ante. The `then` field is the Ast of then branch, and the `otherwise` field is optional (currently) because it holds the Ast of the else branch which currently isn't required. Ie. it is valid to write `if a then b` with no else. The `then` field isn't optional because `if a else b` is not valid syntax, the then branch must always be present. It is just now we want to wrap the whole branch in this call to `Some`. So the branch is still there, there's just an extra function call in it now.
What we could actually do is the reverse - since we'd now always be desugaring `if a then b` to `if a then Some b else None`, we now always have an else branch in every if. So the `otherwise` field is not optional anymore.

jfecher:
I've implemented this in the branch `if-sugar` and now am not sure if the feature is wanted or not. It causes issues with some common patterns such as:
```hs
loop () ->
    if c then
        ...
        recur ()
```
Seen in e.g. `print_string` and several other functions in the prelude alone. When desugared, the if changes to `if c then ... Some (recur ()) else None` which is problematic since if we say the loop returns a type `a` then `Some (recur ())` of type `Maybe a` is returned, we then have `a = Maybe a` which is a type error that can easily confuse users since it is hidden in the syntax sugar.

PR #144, opened 2022-12-03
Fix compile error in prelude/input + rename it

DestyNova:
I ran into errors like this when trying to call `input` in the Prelude:

```
error: No impl found for Extract (Ptr char) usz (ref char)
    buffer#len := '\0'
```

This PR does the following:

* fixes the compile error by replacing `buffer#len` assignment in prelude/input with call to `array_insert` (with help from Ante's Discord chat -- thanks!)
* splits `input` into two functions: `input` and `read_line`, since you'll often want to read a line from stdin without printing a message every time
* adds a check for EOF so programs don't hang forever if empty streams are piped in

### DIFF
--- a/stdlib/prelude.an
+++ b/stdlib/prelude.an
@@ -679,9 +679,9 @@ read_char () : Maybe char =
         then None
         else Some (transmute c : char)
 
-// Read a single line from stdin
-read_line () : string =
-    buffer, len = loop (buffer = null ()) (len = 0) ->
+// Read a single line from stdin, returning (Some s) if we got at least 1 character or None otherwise
+read_line () : Maybe string =
+    line = loop (buffer = null ()) (len = 0) ->
         buffer2 = if len % 32 == 0
             then realloc buffer (len + 32)
             else buffer
@@ -692,13 +692,18 @@ read_line () : string =
 
                 if ch != '\n'
                     then recur buffer2 (len + 1)
-                    else buffer2, len
+                    else Some (buffer2, len)
 
-            | _ -> buffer2, len
+            | _ -> if len > 0
+                        then Some (buffer2, len)
+                        else None
 
 
-    array_insert buffer len '\0'
-    string buffer len
+    match line
+        | Some (buffer, len) ->
+            array_insert buffer len '\0'
+            Some (string buffer len)
+        | _ -> None
 
 /// Print a message then return user input
 input msg : string =
--- a/stdlib/prelude.an
+++ b/stdlib/prelude.an
@@ -675,7 +675,7 @@ next_line (f: InFile) : string =
     string (@cstr) len
 
 // Read a single line from stdin
-read_line : string =
+read_line () : string =
     buffer, len = loop (buffer = null ()) (len = 0) ->
         c = getchar ()
         buffer2 = if len % 32 == 0
@@ -692,6 +692,6 @@ read_line : string =
     string buffer len
 
 /// Print a message then return user input
-input_with_prompt msg : string =
+input msg : string =
     printne msg
     read_line
--- a/stdlib/prelude.an
+++ b/stdlib/prelude.an
@@ -11,7 +11,7 @@ extern
     // printf: Ptr char - ... -> unit
     puts: Ptr char -> i32
     putchar: char -> unit
-    getchar: unit -> char
+    getchar: unit -> i32
     exit: i32 -> never_returns
     malloc: usz -> Ptr a
     calloc: /*items:*/usz - /*size:*/usz -> Ptr a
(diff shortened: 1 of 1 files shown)


stdlib/prelude.an
```diff
@@ -674,23 +674,24 @@ next_line (f: InFile) : string =
...
     string (@cstr) len
 
-
-
-/// Print a message then return user input
-input msg : string =
-    printne msg
-
+// Read a single line from stdin
+read_line : string =
```
jfecher, line 678: Looks like you made a variable here instead of a function, you'll have to add a parameter like so: `read_line () : string = ...`

stdlib/prelude.an
```diff
@@ -674,23 +674,24 @@ next_line (f: InFile) : string =
...
         if c != '\n'
         then recur buffer2 (len + 1)
         else buffer2, len
 
-    buffer#len := '\0'
+    array_insert buffer len '\0'
     string buffer len
+
+/// Print a message then return user input
+input_with_prompt msg : string =
```
jfecher, line 695: Can we keep the old `input` name? It matches python as well.

DestyNova: Sounds fair :)

stdlib/prelude.an
DestyNova: Oops! Guess that means it would have only worked exactly once. Fixed now I hope.

DestyNova:
I see some test failures in CI but not sure how that could happen... will take a look :thinking:

stdlib/prelude.an
```diff
@@ -674,23 +674,24 @@ next_line (f: InFile) : string =
...
         c = getchar ()
         buffer2 = if len % 32 == 0
             then realloc buffer (len + 32)
             else buffer
 
-        buffer2#len := c
+        array_insert buffer2 len c
 
-        if c != '\n'
+        if c != '\n' and (transmute c : i8) >= 0
```
jfecher, line 687: I believe EOF should actually be 255, where 0 is the NUL character (there is also a literal for it: `'\0'`). `next_line` checks for 255 as well for example. You can verify its working by calling the read_line function and hitting Ctrl+D after typing some text to issue the EOF character normally. I tried changing it to this locally and it does seem like I need to hit Ctrl+D twice for some reason.

DestyNova: Oh, I see what you mean. EOF is denoted by `-1` when calling `getchar()` in C, since it [returns an int](https://stackoverflow.com/a/19716188/2161072) rather than an unsigned char. But in Ante's prelude, we declare it as `unit -> char` so it's downcasted to 8 bits and turns into 255 due to the 2's complement representation.

Normal input can contain ASCII value 255 without intending to end the input, so... maybe we should change `getchar` to match C semantics?

DestyNova: @jfecher how about the latest change, declaring `getchar` to return int, and adding `read_char` to return a `Maybe char` taking EOF into account? I _think_ the logic in `read_line` is still correct (although I managed to produce a segfault when returning the old buffer by accident, should be good now).

jfecher: Sounds good, though it doesn't look like the latest change you mentioned has been pushed

DestyNova: Wow, I really do need sleep :laughing: Pushed now.

DestyNova: Hmm, I wonder if `read_line` should also return `Maybe string`. Otherwise you probably have to keep manually checking for EOF (some other way...).

DestyNova: Oops... I changed it to do that, got it working in one program, but failing in another for some reason I don't understand:

```
thread 'main' panicked at '- inst0: The bitcast argument v0 doesn't fit in a type of 8 bits
', src/cranelift_backend/context.rs:194:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:142:14
   2: core::panicking::panic_display
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:72:5
   3: ante::cranelift_backend::context::Context::codegen_function_body
             at /home/omf/code/ante/compiler/src/cranelift_backend/context.rs:194:13
   4: ante::cranelift_backend::context::Context::codegen_all
             at /home/omf/code/ante/compiler/src/cranelift_backend/context.rs:152:13
   5: ante::cranelift_backend::run
             at /home/omf/code/ante/compiler/src/cranelift_backend/mod.rs:23:5
   6: ante::compile
             at /home/omf/code/ante/compiler/src/main.rs:171:9
   7: ante::main
             at /home/omf/code/ante/compiler/src/main.rs:100:9
   8: core::ops::function::FnOnce::call_once
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/ops/function.rs:248:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

DestyNova: (With the following one-line test program: `s = read_line ()`)

DestyNova: Interestingly, it compiles just fine if I add `-O3` :thinking:

jfecher (commented):
Tested locally and am running into the instruction error as well, though have yet to pinpoint its origin. It looks like a case where we're generated bad IR and cranelift is more strict about it than llvm - hence why it works with `-O3` which implicitly switches to the llvm backend since cranelift does not do much optimization. I'll try to look at it more in depth tomorrow but I'd like to avoid merging this until I can fix the error.

jfecher:
Looks like the original error from not using `array_insert` has been fixed. I've also split input into input and read_line as well. I'm closing this since the original error is gone now, but will open a new issue for the bitcast panic if I can reproduce it again.

PR #145, opened 2023-01-01
Clean up CLI help messages

ehllie:
Currently invoking just the command `ante` brings up a somewhat confusing message:
```console
error: The following required arguments were not provided:
  <--shell-completion <SHELL_COMPLETION>|FILE>

Usage: ante <--shell-completion <SHELL_COMPLETION>|FILE>

For more information try '--help'
```
The information on how to generate shell_completions is not very relevant for the end user, as that is usually done during installation. I've moved the shell completion option to a separate parser which will hide it from both the generated completion scripts, and error and help messages for the cli.
```console
error: The following required arguments were not provided:
  <FILE>

Usage: ante <FILE>

For more information try '--help'
```

I've also added rust lsp tooling to the devShell, so that the dev environment has a version of them compiled with the same version of rustc as is being used for ante.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,7 +15,8 @@
           inherit system;
           overlays = [ self.overlays.default ];
         };
-        inherit (pkgs) ante mkShell;
+        inherit (pkgs)
+          ante mkShell rust-analyzer clippy rustfmt;
       in
       {
         packages = {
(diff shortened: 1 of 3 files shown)


jfecher (approved):
Looks great, thanks for being on top of this! Definitely a lot less confusing for users.

The only tests that are failing are the 3 that are failing on master as well so we can safely merge this.

PR #157, opened 2023-02-15
Update nix derivation

ehllie:
Pull request https://github.com/jfecher/ante/commit/fa2e464fb68aa6da1f1252daf14555b0b334dce3 updated `Cargo.lock`. With it came a dependency on `rustc` version `1.14` or higher. I've updated the vendored cargo dependency hash to address the former, and `flake.lock` to use a newer rev of `nixpkgs` to address the latter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
 /target
 .vscode
+
+# Nix build artifacts
+/result
--- a/default.nix
+++ b/default.nix
@@ -20,7 +20,7 @@ let
       pname = "ante";
       src = ./.;
       inherit (toml.package) version;
-      cargoSha256 = "CsWiTDzXMWtnbNBLVSxH0YmXez3YZ+i/W+uPcpgIRDo=";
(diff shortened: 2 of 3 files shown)


jfecher:
Thank you :)

I wonder if there is some way to automate this change or at least detect the need for it in the future.

jfecher (approved):

ehllie:
We only need to include a cargo hash in the derivation expression because of inkwell being a git dependency. If all dependencies came from crates.io, nix would have been able to pre compute the dependencies using just the `Cargo.lock`, meaning no need for that manual update step each time `Cargo.lock` changes. Coincidentally, inkwell seems to have recently updated itself on there, so we might be able use the version from there.

A check can be done with a github action that tries to build the derivation. It would show up as a failed build then. That could also include uploading the artifacts to a binary cache like cachix, so that people installing the compiler using nix wouldn't need to compile ante themselves.

Should I create issues for those? Doing the former should be trivial, and the latter might take some messing about with github actions and setting up cachix tokens, but shouldn't be too difficult either.

PR #160, opened 2023-02-21
Various fixes

iacore:
See commit log for details

### DIFF
--- a/src/cranelift_backend/module.rs
+++ b/src/cranelift_backend/module.rs
@@ -40,10 +40,19 @@ impl DynModule {
         }
     }
 
-    pub fn finish(self, main_id: FuncId, output_file: &Path) {
+    pub fn finish(&mut self) {
         match self {
-            DynModule::Jit(mut module) => {
+            DynModule::Jit(module) => {
                 module.finalize_definitions();
+            },
+            DynModule::Static(_module) => {
+            },
+        }
+    }
+
+    pub fn run(self, main_id: FuncId, output_file: &Path) {
+        match self {
+            DynModule::Jit(module) => {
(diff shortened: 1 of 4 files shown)


src/cranelift_backend/module.rs
```diff
@@ -40,10 +40,19 @@ impl DynModule {
         }
     }
 
-    pub fn finish(self, main_id: FuncId, output_file: &Path) {
+    pub fn finish(&mut self) {
         match self {
-            DynModule::Jit(mut module) => {
+            DynModule::Jit(module) => {
                 module.finalize_definitions();
+            },
```
jfecher, line 47: Lets change this match to an `if let`

src/main.rs
```diff
@@ -166,13 +166,19 @@ fn compile(args: Cli) {
...
     // Phase 6: Codegen
-    if args.opt_level == '0' && args.backend != Some(Backend::Llvm) {
-        cranelift_backend::run(filename, hir, &args);
-    } else if cfg!(feature = "llvm") {
-        #[cfg(feature = "llvm")]
-        llvm::run(filename, hir, &args);
-    } else {
-        eprintln!("The llvm backend is required for non-debug builds. Recompile ante with --features 'llvm' to enable optimized builds.");
+    let backend =
+        args.backend.unwrap_or_else(|| if args.opt_level == '0' { Backend::Cranelift } else { Backend::Llvm });
```
jfecher, line 170: This reads somewhat worse to me to be honest, I think we should revert it. Perhaps it can be revisited if there is ever a third backend or the logic becomes more complex

jfecher (approved):
I seem to have forgotten about this PR, sorry about that. Lets merge it.

Issue #163, opened 2023-09-01
Implement Named Constructor syntax


jfecher:
Ante should be able to construct struct values while naming each field in any order. The syntax should be as follows:

```ocaml
MyStruct with
    field1 = expr1
    field2 = expr2
    ...
    fieldN = exprN
```

If no indent is provided, fields can also be given on the same line, separated by commas:

```ocaml
MyStruct with field1 = expr1, field2 = expr2, ..., fieldN = exprN
```

Note that in this form, constructing tuple values will require parenthesis. Expressions can also be omitted if they are syntactically a variable with the same name as the field. That is, the following:

```ocaml
Foo with bar, baz
```

Is equivalent to 

```ocaml
Foo with bar = bar, baz = baz
```

jfecher:
If a contributor wants to work on this issue, feel free to tackle it in chunks. E.g. adding parser support first. Then in a later PR add name resolution support. During name resolution, we can desugar to a normal constructor expression with positional fields so that no later pass will need to handle these expressions.

```ocaml
MyStruct with field1 = expr1, ..., fieldN = exprN
```
Can be desugared to
```ocaml
_v1 = expr1
_v2 = expr2
MyStruct _v1 _v2
```
Where an outer block expression holds variable definitions for each field to preserve order of operations in case the fields need to be rearranged for the constructor expression.

jfecher:
The remaining work on this issue is to implement name resolution, type checking, and monomorphization for named constructors. It's likely that after name resolution we'll be able to desugar named constructors into regular constructors, e.g:
```hs
MyConstructor with
    field3 = expr3
    field2 = expr2
    field1 = expr1
```
Will desugar to:
```rs
// Fields are brought out into Definitions to preserve evaluation order.
// They're given a name guaranteed not to clash with any identifier in the source program.
$field3 = expr3
$field2 = expr2
$field1 = expr1
MyConstructor $field1 $field2 $field3
```

jfecher:
Implemented in https://github.com/jfecher/ante/commit/b51e02723a5cfb8eb9f98c39f2c446ab86dcbbc8

PR #164, opened 2023-09-02
Add named constructor parsing

nrabulinski:
Addresses the parsing part of #163.
I recommend looking only at the second commit, the first is just noise. Feel free to cherry-pick only the second one. I included it because `rustfmt` is automatically applied for me.

I tried to cover all the edge cases in `examples/parsing/named_constructor.an` but if I missed or misunderstood something feel free to point it out.

### DIFF
--- a/examples/parsing/named_constructor.an
+++ b/examples/parsing/named_constructor.an
@@ -1,0 +1,50 @@
+// Explicit names
+T with b = "one", a = 1
+
+T with
+    b = "one"
+    a = 1
+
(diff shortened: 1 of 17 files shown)


jfecher (approved):
Looks like a good start! I'd eventually like to be able to parse more expressions then just function arguments in the rhs of the `=` of each field, but I can see how this would be difficult with the expression parser trying to parse commas as tuples normally.

Thanks for the PR!

PR #167, opened 2023-10-18
Switch to flake-parts and nci

ehllie:
Switched the nix expression used for building ante to use [flake-parts](https://github.com/hercules-ci/flake-parts) and [nci](https://github.com/yusdacra/nix-cargo-integration) instead of manually calling `buildRustPackage`. It should lower the amount of maintenance required for it, since the dependency hashes won't need to be updated each time `Cargo.lock` changes like it did recently as those are not required by `nci` at all. The `flake.nix` is also much simpler now thanks to `flake-parts`.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/crates.nix
+++ b/crates.nix
@@ -1,0 +1,52 @@
+{ ... }: {
+  perSystem = { pkgs, lib, ... }: {
+    nci =
+      let
+        llvmPackages = pkgs.llvmPackages_16;
+        major = lib.versions.major llvmPackages.llvm.version;
+        minor = lib.versions.minor llvmPackages.llvm.version;
+        llvm-sys-ver = "${major}${builtins.substring 0 1 minor}";
+        env = {
+          "LLVM_SYS_${llvm-sys-ver}_PREFIX" = llvmPackages.llvm.dev;
+        };
+      in
+      {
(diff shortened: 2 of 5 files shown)


jfecher (approved):
Looks good, thank you :slightly_smiling_face:

jfecher:
@ehllie, I just pushed another commit updating some dependencies. Will this break the nix build?

ehllie:
> @ehllie, I just pushed another commit updating some dependencies. Will this break the nix build?

No, the nix build still works fine. It should be much more resilient now.
```console
~
❯ nix shell github:jfecher/ante

~
❯ file $(realpath $(which ante))
/nix/store/ilipmv42z4ms0rqi1ddh4fhv25vi1dw8-ante-0.1.1/bin/ante: Mach-O 64-bit executable arm64

~
❯ ante
error: the following required arguments were not provided:
  <FILE>

Usage: ante <FILE>

For more information, try '--help'.

~
❯ exit

~ took 7m36s
❯
```
One extra thing could be adding a github action for building the binary and uploading it to a binary cache like cachix so that it wouldn't take ~7 minutes to compile, but instead get downloaded from the substituter. It's up to you if you think that's worthwhile. I could look into setting it up if you want.

jfecher:
@ehllie I think we should be good for now, but thank you. It may be something to revisit once the compiler is more mature or if it ever becomes more widely used though.

Issue #169, opened 2023-12-08
Implement `can Effect` syntax for annotating effects


jfecher:
The `effects3` branch is currently under development for implementing effects in ante's backend. It is expected that the type system component of effects is already complete but a crucial part is missing - the ability to annotate the expected effects a function has.

A function should be able to specify its effects via a new `can` keyword:
```hs
foo (x: I32) : I32 can Effect1, Effect2 =
    effect1 x
    effect2 x
```
If a function's return type is inferred, the function's effects are also inferred:
```hs
foo (x: I32) =
    effect1 x
    effect2 x
```
If a function's return type is specified but no effects are specified, we should assume the function is meant to be pure rather than inferring the effects:
```hs
foo (x: I32) : I32 =
    // Error: effect Effect1 is unhandled
    effect1 x
    // Error: effect Effect2 is unhandled
    effect2 x
```

jfecher:
Parsing for this issue has been implemented by https://github.com/jfecher/ante/pull/175. Next up is integration into name resolution and type checking for these annotations. E.g. name resolution should ensure the effects are declared names, and type inference can check this set against the inferred set to issue an error if the inferred set is not a subset of the declared set.

Open question: Should ante issue a warning during type inference if the inferred effect set is a subset of the declared set instead of equal to it? My current thinking is that this would discourage use of effect aliases like `IO` which contain several effects in them that may be unused.

PR #175, opened 2023-12-24
Extend the parser to accept effect annotations.

eldesh:
#169
Extend the parser to accept effect annotations.

Here is an example of what it will be able to accept.

```ante
foo (f: Unit -> a can GiveInt) : a = ...

foo () : Unit can Throw FileError, Throw ParseError, Throw BarError = ...

foo () can Throw FileError, Throw ParseError, Throw BarError = ...
```

No semantic analysis is done, just parsing.


examples/parsing/effects.an
```diff
@@ -24,6 +24,23 @@ handle ()
 | return _ -> "foo"
 | log _ -> "impossible"
```
eldesh, line 26: ported from language tour.

src/parser/ast.rs
```diff
@@ -205,13 +206,19 @@ pub enum Type<'a> {
     Boolean(Location<'a>),
     Unit(Location<'a>),
     Reference(Location<'a>),
-    Function(Vec<Type<'a>>, Box<Type<'a>>, /*varargs:*/ bool, /*closure*/ bool, Location<'a>),
+    Function(Vec<Type<'a>>, Box<Type<'a>>, Vec<Effect<'a>>, /*varargs:*/ bool, /*closure*/ bool, Location<'a>),
```
eldesh, line 209: Is non-ordered container preferable?

jfecher: Likely, but this is fine as a first pass. The order of effect annotations doesn't affect anything, but the order of effect handlers in the code can affect the output.

src/parser/mod.rs
```diff
@@ -116,6 +117,8 @@ parser!(varargs location -> 'b () =
     ()
 );
 
+// Parses the return type of function.
+// This may involve effects.
```
jfecher, line 121:
```suggestion
```
This comment seems outdated seeing as the `function_return_type` parser is separate from the `effect_set` parser in `function_definition`

src/parser/pretty_printer.rs
```diff
@@ -53,6 +53,9 @@ impl<'a> Display for ast::Lambda<'a> {
         if let Some(typ) = &self.return_type {
             write!(f, " : {}", typ)?;
         }
+        if self.effects.len() > 0 {
+          write!(f, " can {}", join_with(&self.effects, ","))?;
```
jfecher, line 57:
```suggestion
          write!(f, " can {}", join_with(&self.effects, ", "))?;
```

jfecher (approved):
Looks good, thanks!

I think one change I'll add in a separate commit is to have the effect_set parser return an `Option<Vec<...>>` instead of just a `Vec<...>` so that we can differentiate between "user did not specify an effect set" and "user specified there should be no effects." Thanks for the help on this issue!

Issue #179, opened 2024-01-14
Create a LSP server for Ante


jfecher:
Ante needs a Language Server Protocol server to give errors and warnings as users write their programs in LSP compatible editors.

This is obviously quite a large task so it can be broken up into as many smaller tasks as desired. Here's some examples:
- [x] Create a LSP server skeleton that runs whenever Ante code is opened in an editor
- [x] Show errors & warnings when a file is saved
- [x] Implement go to definition
- [ ] Implement go to declaration
- [x] Implement "hover" action to show a variable's type.

To limit scope somewhat, I can help with any changes needed in the compiler. I also think the LSP's design should start as simple as possible, e.g. only updating when a file is saved to avoid worrying about partially correct programs for now. Ante's current parser is not recoverable so it will not work well on partial files.

jfecher:
@ehllie I think a separate repository might be best. I'm absolutely in favor of using any libraries for the LSP to make things easier. You can take a look at https://github.com/jfecher/ante-lsp for a possible good starting point (and place to merge things into), it uses tower-lsp and was forked from the boilerplate repo. I haven't looked at it in a while though so I doubt it's in a working state.

ehllie:
I've noticed a couple things that would need to be done for the language server to be effective:

- Make name resolution and type checking steps produce a list of errors, warnings and notes rather than printing them directly to stderr or panicking.
- Make the AST not require a lifetime parameter. The lifetime of AST being tied to the lifetime of the Path reference makes it awkward to work with in a language server context, since the filename is created from the request, which only lives inside the body of a function that receives it. If we want the AST to live longer in the cache it needs to take ownership of the path.
- Fix the `unsafe impl Send for ...` lines? They are marked with a comment saying `TODO: Remove`, and it's usually not the best idea to lie to the compiler 😛. Not sure how much of an issue it could be though.

jfecher:
I'm going to close this issue since most items are completed. I'll break out go to declaration into another issue

PR #182, opened 2024-01-18
Basic LSP implementation 

ehllie:
#179
I've set up the language server as a workspace for now since it makes working on the codebase easier that way. I did not copy the entire boilerplate project, but rather am just using them as reference and copying and pasting when necessary.

So far I've managed to get parsing errors displayed, and it was surprisingly straightforward

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.21.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "ahash"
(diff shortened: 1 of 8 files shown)


jfecher (commented):
This looks amazing, thank you!

Is there any support you need from me to get a minimal version merged soon? As long as it compiles we can merge this in parts - we don't need to wait until the whole LSP is done.

I'm away from my computer ATM so I've yet to test this yet.

ehllie:
I haven't done any work to make this a valid vscode extension like the template project does. Not sure if it would be in scope for the initial pr, but it would definitely make testing the server easier for people that do use it as their primary editor.

Other than that, I might need to change some `String` to `Arc<str>` since things like the `Ast`, and the references to paths it has need to outlive the request parameters, and it's a little tricky to do currently with those things only being borrowed.

ehllie (commented):
So this commit is a little bit silly. Removing the lifetime from the `Location` type propagated out to almost every part of the codebase

ehllie:
This should do be a good first step. Doing a massive rewrite and removing the lifetime parameters from `Location` and as a result `Ast` as well would probably be best left for a separate PR

jfecher:
@ehllie I wonder if we need to get rid of those lifetime parameters or if we can instead keep them and avoid storing the Ast in the language server.

> I haven't done any work to make this a valid vscode extension like the template project does.

That's fine, I think a vscode extension should be a separate repo from the language server anyway. We also don't need to worry about other users testing until we get the basics down first. Since it sounds like we both use (neo)vim we can just focus on that for the moment.

jfecher (approved):
LGTM, we can use this as a base to build upon more features. Thank you!

PR #184, opened 2024-01-24
Typechecker diagnostic messages in language server

ehllie:
Further work towards #179


ehllie:
I've left some relevant comments. This should be good to merge as is, but I have one idea/ suggestion for a change to the compiler that would help with the language server. It would be something akin to passing a file handler to the compiler and cache, and rather than reading files by creating a `File`, it would ask the handler for the file contents. That way the lsp could provide access to files that first goes through the `document_map`, allowing the compiler to run even if the edits made to the files are not saved to the disk, and also possibly speed up compilation by skipping the need to do filesystem io each time.

ante-ls/src/main.rs
```diff
@@ -1,34 +1,48 @@
...
 #[derive(Debug)]
 struct Backend {
     client: Client,
     document_map: DashMap<Url, Rope>,
 }
 
-/// Lets you skip writing the `Type::` part in `..Type::default()` calls
-fn default<T: Default>() -> T {
-    T::default()
-}
```
jfecher, line 20: Thanks for removing this

ante-ls/src/main.rs
```diff
@@ -112,51 +145,110 @@ fn rope_range_to_lsp_range(range: std::ops::Range<usize>, rope: &Rope) -> Range
...
     async fn update_diagnostics(&self, uri: Url, rope: &Rope) {
+        let root = match current_dir() {
+            Ok(root) => root,
+            Err(_) => {
+                self.client.log_message(MessageType::ERROR, "Failed to get current directory".to_string()).await;
+                return;
+            },
+        };
+        // We want the filename to be relative to the root for nicer error messages.
+        // This could fail on windows when the root is on a different drive than the file.
```
jfecher, line 158: I suppose one instance this can happen is when using libraries stored on different drives

jfecher (approved):
image
Amazing 🤩!
Next steps would be to remove the full file path and source line from the error message I think. If we have access to the original Diagnostic, the inner DiagnosticKind should give us just the error message itself.

I agree with your suggestions you gave as well, it'd be good if we could retrieve the file contents directly from the ls and didn't have to re-read them.

It'd also be nice if we were able to make use of the `check` method in ante/src/main.rs so we don't have to repeat logic for lexing, parsing, name resolution, and type checking in the language server.

ehllie:
> image Amazing 🤩! Next steps would be to remove the full file path and source line from the error message I think. If we have access to the original Diagnostic, the inner DiagnosticKind should give us just the error message itself.
> 
> I agree with your suggestions you gave as well, it'd be good if we could retrieve the file contents directly from the ls and didn't have to re-read them.
> 
> It'd also be nice if we were able to make use of the `check` method in ante/src/main.rs so we don't have to repeat logic for lexing, parsing, name resolution, and type checking in the language server.

It's interesting how it's using absolute paths for you. It *should* be using relative paths to the root of the project, and at least it's doing that for me. But I guess I'm currently doing it by "lying" to the compiler about what path the file is at, and it could be done in a smarter way.
<details>

image
</details>

PR #185, opened 2024-01-25
Update flake to include the language server #179

ehllie:
Updated the flake to also provide an `ante-ls` output and did some basic cleaning up of the existing package outputs. We might eventually want to move the language server to a separate repository, but in the meantime there's no harm in having the main flake provide it as well.


jfecher (approved):
👍

PR #186, opened 2024-01-26
Add a file cache

ehllie:
I've added a file cache to ModuleCache. That should allow for passing in memory buffers from the lsp to the compiler, and therefore updating diagnostics without having to wait for a file write. It also changes error messages to be relative paths from the root each module is in, and as such the expected errors for tests had to be updated.

Additionally, I've separated the check function into a frontend module that is reused in the language server like previously suggested.

### DIFF
--- a/ante-ls/src/main.rs
+++ b/ante-ls/src/main.rs
@@ -7,10 +7,7 @@ use std::{
 use ante::{
     cache::{cached_read, ModuleCache},
     error::{location::Locatable, ErrorType},
-    lexer::Lexer,
-    nameresolution::NameResolver,
-    parser::parse,
-    types::typechecker,
+    frontend,
 };
 
 use dashmap::DashMap;
@@ -141,19 +138,7 @@ impl Backend {
         let file_cache = paths.iter().zip(contents.into_iter()).map(|(p, c)| (p.as_path(), c)).collect();
         let mut cache = ModuleCache::new(cache_root, file_cache);
 
-        let tokens = Lexer::new(filename, &rope.to_string()).collect::<Vec<_>>();
-        match parse(&tokens) {
-            Ok(ast) => {
-                NameResolver::start(ast, &mut cache);
-                if cache.error_count() == 0 {
-                    let ast = cache.parse_trees.get_mut(0).unwrap();
-                    typechecker::infer_ast(ast, &mut cache);
-                }
-            },
-            Err(err) => {
-                cache.push_full_diagnostic(err.into_diagnostic());
-            },
-        };
+        let _ = frontend::check(filename, rope.to_string(), &mut cache, frontend::FrontendPhase::TypeCheck, false);
 
         // Diagnostics for a document get cleared only when an empty list is sent for it's Uri.
         // This presents an issue, as when we have files A and B, where file A imports the file B,
--- a/ante-ls/src/main.rs
+++ b/ante-ls/src/main.rs
@@ -1,11 +1,11 @@
 use std::{
     collections::HashMap,
-    env::{current_dir, set_current_dir},
+    env::set_current_dir,
     path::{Path, PathBuf},
 };
 
 use ante::{
-    cache::ModuleCache,
+    cache::{cached_read, ModuleCache},
     error::{location::Locatable, ErrorType},
     lexer::Lexer,
     nameresolution::NameResolver,
@@ -73,9 +73,6 @@ impl LanguageServer for Backend {
         self.update_diagnostics(params.text_document.uri, &rope).await;
     }
 
-    // The diagnostics can't be updated on change, because the content of the file in the file system
-    // is not guaranteed to be the same as the content of the file in the editor. This will result in
-    // a panic when running Diagnostic::format, and the column are lengths different than expected.
     async fn did_change(&self, params: DidChangeTextDocumentParams) {
         self.client.log_message(MessageType::LOG, format!("ante_ls did_change: {:?}", params)).await;
         self.document_map.alter(&params.text_document.uri, |_, mut rope| {
@@ -90,25 +87,9 @@ impl LanguageServer for Backend {
             }
             rope
         });
-    }
-}
-
-fn relative_path<P1: AsRef<Path>, P2: AsRef<Path>>(root: P1, path: P2) -> Option<PathBuf> {
-    let path = path.as_ref();
-    if let Ok(path) = path.strip_prefix(&root) {
-        return Some(path.to_path_buf());
-    }
-
-    let mut acc = PathBuf::new();
-    let mut root = root.as_ref();
-    loop {
-        if let Ok(path) = path.strip_prefix(root) {
-            acc.push(path);
-            return Some(acc);
-        } else {
-            acc.push("..");
-            root = root.parent()?;
-        }
+        if let Some(rope) = self.document_map.get(&params.text_document.uri) {
+            self.update_diagnostics(params.text_document.uri, &rope).await;
+        };
     }
 }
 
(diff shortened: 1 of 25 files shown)


ante-ls/src/main.rs
```diff
@@ -196,40 +162,26 @@ impl Backend {
...
-                    let rope = Rope::from_str(&std::fs::read_to_string(uri.path()).unwrap());
-                    self.document_map.insert(uri.clone(), rope.clone());
+                    let contents = cached_read(&cache.file_cache, loc.filename).unwrap();
+                    let rope = Rope::from_str(&contents);
+                    self.document_map.insert(uri.clone(), rope);
                     self.document_map.get(&uri).unwrap()
                 },
             };
 
-            // TODO: This range seems to be off
```
ehllie, line 225: What exactly was off about it? Not sure if it's just fixed now because they seem fine to me:

<details>

image

</details>

jfecher: I have an example I'm working on currently where the location is off:
image

Running the compiler on the file reports:
image

Where the line and column info are correct.

jfecher: You can ignore the contents of the file. I've found a good way of making Ante thread-safe while also allowing aliased mutability without lifetime annotations!

ante-ls/src/main.rs
```diff
@@ -147,32 +125,20 @@ fn rope_range_to_lsp_range(range: std::ops::Range<usize>, rope: &Rope) -> Range
...
-                }
-            },
-            Err(err) => {
-                cache.push_full_diagnostic(err.into_diagnostic());
-            },
-        };
+        let cache_root = filename.parent().unwrap();
+
+        let (paths, contents) =
+            self.document_map.iter().fold((Vec::new(), Vec::new()), |(mut paths, mut contents), item| {
```
jfecher, line 133: I wonder if `unzip` would be cleaner here?

src/cache/mod.rs
```diff
@@ -126,8 +128,13 @@ pub struct ModuleCache<'a> {
 
     /// The number of errors emitted by the program
     pub error_count: usize,
+
+    pub file_cache: FileCache<'a>,
 }
 
+pub type FileCache<'a> = HashMap<&'a Path, String>;
+// type MutCache<'a> = &'a mut HashMap<&'a Path, Box<str>>;
```
jfecher, line 136:
```suggestion
```

src/cache/mod.rs
```diff
@@ -372,8 +379,21 @@ impl<'a> Locatable<'a> for EffectInfo<'a> {
...
+pub fn cached_read(file_cache: &FileCache<'_>, path: &Path) -> Option<String> {
+    file_cache.get(path).cloned().or_else(|| {
+        let file = File::open(path).ok()?;
+        let mut reader = BufReader::new(file);
+        let mut contents = String::new();
+        reader.read_to_string(&mut contents).ok()?;
+
+        Some(contents)
+    })
+}
```
jfecher, line 391: We can avoid cloning a cached File using a `Cow`. Alternatively, maybe we should automatically add a non-cached file's contents to the cache automatically similar to `get_contents`?
```suggestion
pub fn cached_read<'c>(file_cache: &FileCache<'c>, path: &Path) -> Option<Cow<'c, str>> {
    file_cache.get(path).map_or_else(Cow::Borrowed, || {
        let file = File::open(path).ok()?;
        let mut reader = BufReader::new(file);
        let mut contents = String::new();
        reader.read_to_string(&mut contents).ok()?;
        Some(Cow::Owned(contents))
    })
}
```

src/cache/mod.rs
```diff
@@ -417,10 +438,23 @@ impl<'a> ModuleCache<'a> {
...
         }
     }
 
+    pub fn get_contents(&mut self, path: &'a Path) -> Option<String> {
+        let contents = cached_read(&self.file_cache, path)?;
+        if !self.file_cache.contains_key(path) {
+            self.file_cache.insert(path, contents.clone());
+        };
+        Some(contents)
+    }
```
jfecher, line 452:
```suggestion
    pub fn get_contents(&mut self, path: &'a Path) -> Option<&'a str> {
        let has_path = self.file_cache.contains_key(path);
        let contents = cached_read(&self.file_cache, path)?;
        
        if !has_path {
            let contents = contents.into_owned();
            self.file_cache.insert(path, contents);
        }
        
        self.file_cache.get(path).map(|s| s.as_str())
    }
```
A bit awkward to rewrite this to return a borrowed string when possible.

src/error/mod.rs
```diff
@@ -413,18 +412,23 @@ impl<'a> Diagnostic<'a> {
...
         let start = self.location.start;
 
-        writeln!(f, "{}\t{} {}", self.location, self.marker(), self.msg)?;
+        let relative_path =
+            os_agnostic_display_path(cache.strip_root(self.location.filename).unwrap_or(self.location.filename));
 
-        let file_contents = read_file_or_panic(self.location.filename);
+        write!(f, "{}:{}:{}", relative_path, start.line, start.column)?;
+
+        writeln!(f, "\t{} {}", self.marker(), self.msg)?;
```
jfecher, line 429:
```suggestion
        writeln!(f, "{}:{}:{}\t{} {}", relative_path, start.line, start.column, self.marker(), self.msg)?;
```

src/frontend.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+pub enum FrontendResult {
+    Done,
+    ContinueCompilation,
+    Errors,
+}
+
+pub fn check<'a>(
+    filename: &'a Path, main_file_contents: String, cache: &mut ModuleCache<'a>, phase: FrontendPhase, show_time: bool,
```
jfecher, line 24:
```suggestion
    filename: &'a Path, main_file_contents: String, cache: &mut ModuleCache<'a>, stop_after_phase: FrontendPhase, show_time: bool,
```
`phase` sounded a bit too general to me

jfecher (approved):
Really nice! Big improvements in the LS and in getting rid of the read file or panic function while displaying errors.

I think there are some changes we could make to reduce the cloning of files. Since I can't edit this PR myself though, I'll log the cases here, merge this, and make a follow up commit myself to reduce them to reduce your workload.

Thank you again! Do you have plans on what to work on for ante-ls next?

jfecher (approved):

ehllie:
> Really nice! Big improvements in the LS and in getting rid of the read file or panic function while displaying errors.
> 
> I think there are some changes we could make to reduce the cloning of files. Since I can't edit this PR myself though, I'll log the cases here, merge this, and make a follow up commit myself to reduce them to reduce your workload.
> 
> Thank you again! Do you have plans on what to work on for ante-ls next?

Well, one way to reduce the cost of the clones would be using an `Arc`. I tired using an `Rc` at first, but it would not work inside async functions as Rc in not Send + Sync. I figured it's best not to prematurely optimise this, especially since at current scale it seems efficient enough.

I think I'm going to try implementing hover next.

ante-ls/src/main.rs
```diff
@@ -196,40 +162,26 @@ impl Backend {
...
+                    let contents = cached_read(&cache.file_cache, loc.filename).unwrap();
+                    let rope = Rope::from_str(&contents);
+                    self.document_map.insert(uri.clone(), rope);
                     self.document_map.get(&uri).unwrap()
                 },
             };
 
-            // TODO: This range seems to be off
             let range = rope_range_to_lsp_range(loc.start.index..loc.end.index, &rope);
+            let message = format!("{}", diagnostic.display(&cache));
```
jfecher, line 178: Also, it looks like this change on master was accidentally reverted.

jfecher:
> Well, one way to reduce the cost of the clones would be using an Arc. I tired using an Rc at first, but it would not work inside async functions as Rc in not Send + Sync. I figured it's best not to prematurely optimise this, especially since at current scale it seems efficient enough.

Returning a `Cow` works, see 38fab45e145f3b619a38b2af3c08f9d811773c1d

ante-ls/src/main.rs
jfecher: Ok, I've determined this is because of larger unicode characters present in the file causing the internal byte index to be different than the character index.

PR #187, opened 2024-02-07
Lsp hover capability

ehllie:
So far hover only provides any information for variables. Future work could be adding doc comment rendering, as well as support for hovering trait definitions. The language server might also soon be due a slight refactor as the `main.rs` file is starting to grow a little large.

<details>
<summary>Hover preview</summary>

image
</details>


ehllie:
#179

ante-ls/src/main.rs
```diff
@@ -88,58 +91,265 @@ impl LanguageServer for Backend {
...
+    async fn hover(&self, params: HoverParams) -> Result<Option<Hover>> {
+        self.client.log_message(MessageType::LOG, format!("ante-ls hover: {:?}", params)).await;
+        let uri = params.text_document_position_params.text_document.uri;
+        let rope = match self.document_map.get(&uri) {
+            Some(rope) => rope,
+            None => return Ok(None),
+        };
+
+        let cache = self.create_cache(&uri, &rope);
+        let ast = cache.parse_trees.get_mut(0).unwrap();
```
jfecher, line 104: Do we need a map between file names and which parse tree to grab?

ante-ls/src/main.rs
```diff
@@ -88,58 +91,265 @@ impl LanguageServer for Backend {
...
+                    VariableKind::Identifier(name) => name.clone(),
+                    VariableKind::TypeConstructor(name) => name.clone(),
+                    VariableKind::Operator(op) => format!("({})", op),
+                };
+
+                let (typ_str, traits) =
+                    typeprinter::show_type_and_traits(typ, &info.required_traits, &info.trait_info, &cache);
+
+                let value = format!("{} : {}", name, typ_str);
+                let value = if traits.is_empty() { value } else { format!("{}\n  given {}", value, traits.join(", ")) };
```
jfecher, line 131: Since this is meant to be the same as the diagnostics shown with `--show-types`, can you make a wrapper function for these lines so that this part of ante-ls and the `--show-types` flag can share the same formatting function?

ante-ls/src/main.rs
```diff
@@ -88,58 +91,265 @@ impl LanguageServer for Backend {
...
+                let typ = match &info.typ {
+                    Some(typ) => typ,
+                    None => return Ok(None),
+                };
+
+                let name = match &v.kind {
+                    VariableKind::Identifier(name) => name.clone(),
+                    VariableKind::TypeConstructor(name) => name.clone(),
+                    VariableKind::Operator(op) => format!("({})", op),
+                };
```
jfecher, line 125: Hmm, I should probably add a `get_name` function returning a `Cow<str>` as well

jfecher (commented):
Thanks! I have a few questions, then we can get this merged

ante-ls/src/main.rs
ehllie: We're not preserving the cache between each hover, and it get recreated with the file containing the hover location being entry point. As such the 0th parse tree should always contain the file we are interested in.

src/types/typeprinter.rs
ehllie, line 1: I've modified the type printer function itself rather than making a new one, since the only place it was being used other than the language server was the `print_definition_types`.

jfecher (approved):
Great changes, thank you!

It always makes my day each time there's a new PR - and such an exciting one too.

PR #188, opened 2024-02-07
Implement named constructor

ehllie:
Attempts to address #163

### DIFF
--- a/src/nameresolution/mod.rs
+++ b/src/nameresolution/mod.rs
@@ -1627,7 +1627,102 @@ impl<'c> Resolvable<'c> for ast::Handle<'c> {
 impl<'c> Resolvable<'c> for ast::NamedConstructor<'c> {
     fn declare(&mut self, _resolver: &mut NameResolver, _cache: &mut ModuleCache<'c>) {}
 
-    fn define(&mut self, _resolver: &mut NameResolver, _cache: &mut ModuleCache<'c>) {
-        todo!()
+    fn define(&mut self, resolver: &mut NameResolver, cache: &mut ModuleCache<'c>) {
+        let type_name = match self.constructor.as_ref() {
+            Ast::Variable(ast::Variable { kind, .. }) => kind.name(),
+            _ => {
+                // This should never happen since constructor is parsed with the `variant` parser
+                cache.push_diagnostic(
+                    self.constructor.locate(),
+                    D::InternalError("Expected consturctor field to be a Variable"),
+                );
+                return;
+            },
+        };
+        // This will increment the use count for that type.
+        // It will result in it being one higher than it needs to,
+        // as the define pass on the sequence will do it again,
+        // since it ends with a FunctionCall. Is that a problem?
+        let type_info = match resolver.lookup_type(type_name.as_ref(), cache) {
+            Some(id) => &cache.type_infos[id.0],
+            None => {
+                cache.push_diagnostic(self.location, D::NotInScope("Type", type_name.as_ref().clone()));
+                return;
+            },
+        };
+
+        // Field names in the order they appear in the type definition
+        let struct_fields = match &type_info.body {
+            TypeInfoBody::Struct(fields) => fields.iter().map(|field| &field.name),
+            _ => {
+                cache.push_diagnostic(self.constructor.locate(), D::NotAStruct(type_name.as_ref().clone()));
+                return;
+            },
+        };
+        let statements = match self.sequence.as_mut() {
+            Ast::Sequence(ast::Sequence { statements, .. }) => statements,
+            _ => {
+                // This should never happen again, but it's better to emit an error than panic
+                cache.push_diagnostic(self.location, D::InternalError("Expected statements field to be a Variable"));
+                return;
+            },
+        };
+
+        // Fields referenced in the constructor
+        let mut defined_fields = statements
+            .iter()
+            .map(|stmt| {
+                let (variable, location) = match stmt {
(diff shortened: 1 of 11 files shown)


src/nameresolution/mod.rs
```diff
@@ -1627,7 +1627,10 @@ impl<'c> Resolvable<'c> for ast::Handle<'c> {
 impl<'c> Resolvable<'c> for ast::NamedConstructor<'c> {
     fn declare(&mut self, _resolver: &mut NameResolver, _cache: &mut ModuleCache<'c>) {}
 
-    fn define(&mut self, _resolver: &mut NameResolver, _cache: &mut ModuleCache<'c>) {
-        todo!()
+    fn define(&mut self, resolver: &mut NameResolver, cache: &mut ModuleCache<'c>) {
+        self.constructor.define(resolver, cache);
+        for (_, arg_val) in self.args.iter_mut() {
+            arg_val.define(resolver, cache);
+        }
```
jfecher, line 1634: I think you'd save a lot of code if you desugared named constructors into ordinary constructors here. You'd still need to check all fields are present though.

It should desugar to roughly:
```
Sequence(vec![
    Definition(new_var1, field_expr1),
    ...
    Definition(new_varN, field_exprN),
    FunctionCall(Variable(TypeConstructor(Foo)), vec![new_var1, ..., new_varN])
])
```
(where the fields are defined in the order they were specified in the named constructor expr, but applied to the function call in the order Foo's definition expects)
That way we'd get type checking + monomorphization for free

ehllie: Do you mean desugaring into a sequence in the parsing stage, or somehow mutating `self` into a `sequence` in the name resolution stage? Cause I was not certain if we would even have the type info of the struct we're trying to construct at the name resolution stage yet, and as such we would not be able to call the constructor in the right order.

jfecher: I mean mutating `self` during `NameResolution::define` of a named constructor.

Looking at `TypeDefinition::define` though, I suppose we wouldn't necessarily know a types fields until that is called. I suppose we could also do the same mutation but during type checking. Then recursively type check the result.

ehllie: I've decided to just store the inner `Sequence` as a field on `NamedConstructor`, and mutate it during the `define` phase.

jfecher: That's probably a better idea. It'd be tricky trying to mutate a named constructor into a sequence without acces to the actual `&mut Ast` value.

ehllie:
There seems to be a bug in the cranelift backend that I don't fully understand. It compiling the `examples/typechecking/named_constructor.an` file:

```console
$ ante -d --backend llvm examples/typechecking/named_constructor.an
ld: warning: object file (/Users/ellie/Code/ante/examples/typechecking/named_constructor.o) was built for newer macOS version (14.0) than being linked (11.0)

$ ante -d --backend cranelift examples/typechecking/named_constructor.an
thread 'main' panicked at src/cranelift_backend/context.rs:197:13:
- inst0 (v1, v2, v3 = call fn0(v0, v1, v0)): uses value v1 from itself

```

It works while using the llvm backend, so it does not seem to be directly related to this feature, and I remember you mentioning making big changes to the backend soon, so I feel like this PR should be good as is for now.

jfecher:
Hmm, I don't think this PR should affect the backend at all. I wonder if it has an issue that is going unnoticed until it triggers a check in cranelift? I've never seen the "uses value v1 from itself" error before though. Usually it is a type error or using a value before it is defined. I'd double-check the output (specifically the definition ids for each variable) produced by this desugaring by printing out the monomorphized IR.

I believe the only backend changes I had in mind were for the algebraic effects branch so it will still be some time before that is merged.

ehllie:
I had misread you saying that you'd look into the generated IR. The HIR generated is the following
```console
$ ante -e hir examples/typechecking/named_constructor.an

();
x_v0 = ("Hello World", 11_usz);
();
foo_v6 = (hello_foo_v1 42_i32)


hello_foo_v1 = (fn v2 : i32 -> {{Ptr, usz}, i32} =
        (Foo_v5 x_v0 y_v2)
)

Foo_v5 = (fn v3 v4 : {Ptr, usz} -> i32 -> {{Ptr, usz}, i32} = (v3, v4))
```

But generating the cranelift IR panics
```console
$ ante -e ir examples/typechecking/named_constructor.an
main =
function u0:0() -> i32 apple_aarch64 {
    gv0 = symbol colocated u1:0
    sig0 = (i32) -> i64, i64, i32 fast
    fn0 = u0:1 sig0

block0:
    v0 = symbol_value.i64 gv0
    v1 = iconst.i64 11
    v2 = iconst.i32 42
    v3, v4, v5 = call fn0(v2)
    v6 = iconst.i32 0
    return v6
}

lambdav1 =
function u0:1(i32) -> i64, i64, i32 fast {
    sig0 = (i64, i64, i32) -> i64, i64, i32 fast
    fn0 = u0:2 sig0

block0(v0: i32):
    v1, v2, v3 = call fn0(v0, v1, v0)
    return v1, v2, v3
}

thread 'main' panicked at src/cranelift_backend/context.rs:197:13:
- inst0 (v1, v2, v3 = call fn0(v0, v1, v0)): uses value v1 from itself

```

In all honesty, I'm feeling somewhat out of my depth with debugging codegen

jfecher:
@ehllie ah, sorry about that, it has been some time and I just forgot where things stood on this PR. I can take it over and do some debugging if that'd be simpler for you. I'll make the PR on your fork when that is done, sorry for the confusion.

jfecher (approved):
👍

PR #196, opened 2024-02-28
Restore ante-ls goto definition

ehllie:
`ante-ls` built just fine for me before the commit was reverted. It might have had something to do with the awkward git history, since I made a manual merge commit rather than rebasing. It would probably be best to squash down the last couple commits if this builds fine.

### DIFF
--- a/ante-ls/src/main.rs
+++ b/ante-ls/src/main.rs
@@ -15,7 +15,14 @@ use ante::{
 use dashmap::DashMap;
 use futures::future::join_all;
 use ropey::Rope;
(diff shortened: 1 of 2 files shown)


jfecher:
This is building fine for me locally, I'll test again after merging just in case. I have PRs set to always squash & merge so I don't think that was the issue last time.

jfecher:
Ok, master still builds with rustc 1.75

jfecher:
After some investigation https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html#projections-in-opaque-return-types is the feature that is making this work now. So 1.74 is now ante's (or at least ante-ls') minimum rustc version.

jfecher:
Thanks for the PR!

PR #198, opened 2024-04-05
Draft: Add Else Operator

anoojpatel:
Add the Else Operator with type: `(else): Maybe a -> a -> a`

This PR adds a new `Else` binary operator into the AST and handles the tracking of the `Maybe` type constructor from the prelude.

- [x] Extract out `Maybe a` type from prelude
- [x] Handle just the `Maybe a else b` definitions in the AST and Parsing by desugaring into a Match AST
- [ ] Pass Tests

### DIFF
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -132,7 +132,7 @@ pub struct ModuleCache<'a> {
 
     pub file_cache: FileCache,
 
-    pub maybe_type: TypeInfoId,
+    pub maybe_type: Option<TypeInfoId>,
 }
 
 pub type FileCache = HashMap<PathBuf, String>;
@@ -419,7 +419,7 @@ impl<'a> ModuleCache<'a> {
             global_dependency_graph: DependencyGraph::default(),
             diagnostics: Vec::new(),
             error_count: 0,
-            maybe_type: TypeInfoId(0), // sentinel value
+            maybe_type: Some(TypeInfoId(0)), // sentinel value
             file_cache,
         }
     }
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -419,7 +419,7 @@ impl<'a> ModuleCache<'a> {
             global_dependency_graph: DependencyGraph::default(),
             diagnostics: Vec::new(),
             error_count: 0,
-            maybe_type: Some(TypeInfoId(0)), // sentinel value
+            maybe_type: None,
             file_cache,
         }
     }
--- a/src/cache/mod.rs
+++ b/src/cache/mod.rs
@@ -131,6 +131,8 @@ pub struct ModuleCache<'a> {
     pub error_count: usize,
 
     pub file_cache: FileCache,
+
+    pub maybe_type: TypeInfoId,
 }
 
 pub type FileCache = HashMap<PathBuf, String>;
@@ -417,6 +419,7 @@ impl<'a> ModuleCache<'a> {
             global_dependency_graph: DependencyGraph::default(),
             diagnostics: Vec::new(),
             error_count: 0,
+            maybe_type: TypeInfoId(0), // sentinel value
             file_cache,
         }
     }
--- a/src/cranelift_backend/mod.rs
+++ b/src/cranelift_backend/mod.rs
@@ -163,6 +163,33 @@ impl CodeGen for hir::If {
     }
 }
 
+impl CodeGen for hir::Else {
+    fn codegen<'a>(&'a self, context: &mut Context<'a>, builder: &mut FunctionBuilder) -> Value {
+        let expr = builder.create_block();
+        let then_values = context.eval_all_in_block(&self.expr, expr, builder);
+
+        let end = context.new_block_with_arg(&self.result_type, builder);
+
+        let else_none = builder.create_block();
+        builder.ins().jump(else_none, &[]);
+
+        let else_values = context.eval_all_in_block(&self.otherwise, else_none, builder);
+
+        if let Some(else_values) = else_values {
+            builder.ins().jump(end, &else_values);
+        }
+
+        builder.seal_block(end);
+        builder.switch_to_block(end);
+        let end_values = builder.block_params(end);
+        let ret = context.array_to_value(end_values, &self.result_type);
+
+        builder.seal_block(expr);
+        builder.seal_block(else_none);
+        ret
+    }
+}
+
 impl CodeGen for hir::Match {
     fn codegen<'a>(&'a self, context: &mut Context<'a>, builder: &mut FunctionBuilder) -> Value {
         context.codegen_match(self, builder)
--- a/src/cranelift_backend/mod.rs
+++ b/src/cranelift_backend/mod.rs
@@ -165,15 +165,15 @@ impl CodeGen for hir::If {
 
 impl CodeGen for hir::Else {
     fn codegen<'a>(&'a self, context: &mut Context<'a>, builder: &mut FunctionBuilder) -> Value {
-        let expr = builder.create_block();
-        let then_values = context.eval_all_in_block(&self.expr, expr, builder);
+        let lhs = builder.create_block();
+        let then_values = context.eval_all_in_block(&self.lhs, lhs, builder);
 
         let end = context.new_block_with_arg(&self.result_type, builder);
 
         let else_none = builder.create_block();
         builder.ins().jump(else_none, &[]);
 
-        let else_values = context.eval_all_in_block(&self.otherwise, else_none, builder);
+        let else_values = context.eval_all_in_block(&self.rhs, else_none, builder);
 
         if let Some(else_values) = else_values {
             builder.ins().jump(end, &else_values);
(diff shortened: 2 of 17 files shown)


src/types/typechecker.rs
```diff
@@ -1728,6 +1728,28 @@ impl<'a> Inferable<'a> for ast::If<'a> {
     }
 }
 
+impl<'a> Inferable<'a> for ast::Else<'a> {
+    fn infer_impl(&mut self, cache: &mut ModuleCache<'a>) -> TypeResult {
+        // Create new serogate Variable x
+        let x = next_type_variable(cache);
+        //let x = Type::TypeVariable(id);
+        // Construct a Maybe x Type
+        //let maybe_typeinfo_id = cache.type_infos.get(cache.maybe_type.0);
```
jfecher, line 1737:
```suggestion
```
No need to repeat in comments what is already done on the code in the next lines.

Comment-wise, I usually only leave comments to explain  _why_ I'm doing something rather than _what_ I'm doing. For example, if there's a specific edge case that prevents the code from being simpler I'll write that down so I don't try to simplify the code later and break it.

Other than that though, type checking code looks good

src/types/typechecker.rs
```diff
@@ -1728,6 +1728,28 @@ impl<'a> Inferable<'a> for ast::If<'a> {
...
+        let maybe_type = UserDefined(cache.maybe_type);
+        let maybe_x = TypeApplication(Box::new(maybe_type), vec![x.clone()]);
+        // Unify Maybe x with expr and raise if Type is mismatched
+        let mut expr = infer(self.expr.as_mut(), cache);
+        unify(&maybe_x, &mut expr.typ, self.expr.locate(), cache, TE::ElseBranchMismatch);
```
jfecher, line 1742:
```suggestion
        unify(&maybe_x, &expr.typ, self.expr.locate(), cache, TE::ElseBranchMismatch);
```

src/llvm/mod.rs
```diff
@@ -627,6 +627,47 @@ impl<'g> CodeGen<'g> for hir::If {
...
 }
 
+impl<'g> CodeGen<'g> for hir::Else {
+    fn codegen(&self, generator: &mut Generator<'g>) -> BasicValueEnum<'g> {
+        let current_function = generator.current_function();
+        let expr_block = generator.context.append_basic_block(current_function, "expr");
+        let end_block = generator.context.append_basic_block(current_function, "end_else");
+
+        // Setup conditional jump
+        let else_block = generator.context.append_basic_block(current_function, "else");
```
jfecher, line 637: We don't need a new block for `expr` - it can be evaluated in the current block since we don't need to conditionally jump to it or anything.

You can think of `a else b` as desugaring to:
```hs
match a
| Some value -> value
| None -> b
```
Which in llvm is something like the following pseudocode:
```llvm
%0 = ... evaluate a ...
%1 = extract tag value from %0
br i1 %1, label %none_case, label %some_case              (build_conditional_branch)

none_case:
  %2 = ... evaluate b ...
  br label %end
some_case:
  %3 = extract non-tag value from %0
  br label %end
end:
  %4 = phi <result type> [%2, %none_case ], [%3, %some_case]                    result of the match
```

You can always use `ante --emit ir --backend llvm <filename>` to see what llvm you're emitting for a particular file

src/cranelift_backend/mod.rs
```diff
@@ -163,6 +163,33 @@ impl CodeGen for hir::If {
     }
 }
 
+impl CodeGen for hir::Else {
+    fn codegen<'a>(&'a self, context: &mut Context<'a>, builder: &mut FunctionBuilder) -> Value {
+        let expr = builder.create_block();
```
jfecher, line 168: Similar to the llvm code - we don't need a new block for `expr` - it can be evaluated in the current block since we don't need to conditionally jump to it or anything.

You can think of `a else b` as desugaring to:
```hs
match a
| Some value -> value
| None -> b
```
Which in cranelift is something like the following pseudocode:
```llvm
v0, v1, .., vN = ... evaluate a ...     ; cranelift doesn't have structs, so all values are always "unpacked"
                                        ; because of this v0 is the tag value, and the N values after it are the inner payload of the Maybe
brnz v0, some_block    ; if the tag is non-zero, go to some_block

some_block:
  jump end(v1, .., vN)  ; jump to end with the arguments v1 to vN (the inner Maybe values)

none_case:
  v2, .., v2N = ... evaluate b ...
  jump end(v2, .., v2N)      ; You should get N values from evaluating b as well since it's the same type as the inner maybe values above

end(v3, .., v3N):   ; results of the match are block arguments (cranelift uses these instead of Phi)
  ... block continues ...
```

You can always use `ante --emit ir --backend craneilft <filename>` to see what cranelift you're emitting for a particular file

src/hir/mod.rs
```diff
@@ -108,6 +108,13 @@ pub struct If {
     pub result_type: Type,
 }
 
+#[derive(Debug, Clone)]
+pub struct Else {
+    pub expr: Box<Ast>,
+    pub otherwise: Box<Ast>,
+    pub result_type: Type,
```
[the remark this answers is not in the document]
anoojpatel: Done!

src/llvm/mod.rs
anoojpatel: That makes sense! One thing I'm trying to learn is how exactly to extract out the tag and non-tag values from the codegen'd IR? Is there a specific section of the llvm codegen I could potentially pattern match off?

jfecher: Yeah, you can look at the codegen portions for decision trees which is what match expressions are compiled to. It can be a bit tricky though since it depends on type information. Another option would be desugaring the Else nodes into Match nodes during monomorphisation most likely. Then the `hir` wouldn't need Else nodes at all.

src/cache/mod.rs
```diff
@@ -419,7 +419,7 @@ impl<'a> ModuleCache<'a> {
             global_dependency_graph: DependencyGraph::default(),
             diagnostics: Vec::new(),
             error_count: 0,
-            maybe_type: TypeInfoId(0), // sentinel value
+            maybe_type: Some(TypeInfoId(0)), // sentinel value
```
jfecher, line 422:
```suggestion
            maybe_type: None,
```

src/hir/printer.rs
```diff
@@ -179,11 +179,11 @@ impl FmtAst for If {
...
-        write!(f, "Maybe a")?;
-        printer.block(self.expr.as_ref(), f)?;
-        write!(f, "else ")?;
-        printer.block(self.otherwise.as_ref(), f)?;
-        write!(f, " endelse")
+        write!(f, "(")?;
+        printer.block(self.lhs.as_ref(), f)?;
+        write!(f, " else ")?;
+        printer.block(self.rhs.as_ref(), f)?;
+        write!(f, ") ")
```
jfecher, line 186:
```suggestion
        write!(f, ")")
```

src/parser/mod.rs
```diff
@@ -633,6 +633,14 @@ parser!(match_branch _loc -> 'b (Ast<'b>, Ast<'b>) =
     (pattern, branch)
 );
 
+parser!(without_else_expr _loc =
+    _ <- maybe_newline;
+    expr <- block_or_statement;
+    _ <- expect(Token::Else);
+    otherwise !<- block_or_statement;
+    otherwise
+);
```
jfecher, line 642: The `if_expr` parser above already parses a `if <block_or_statement> then <block_or_statement>` before this so using this parser afterward would parse:
```rs
if <block_or_statement>
then <block_or_statement>
<block_or_statement> else <block_or_statement>
```
We don't need to change the `if_expr` parser for this parser, it should continue to parse `if a then b else c`, the `else` expression added in this PR can be a separate binary operator expression entirely.

src/cranelift_backend/mod.rs
```diff
@@ -165,15 +165,15 @@ impl CodeGen for hir::If {
...
+        let lhs = builder.create_block();
+        let then_values = context.eval_all_in_block(&self.lhs, lhs, builder);
 
         let end = context.new_block_with_arg(&self.result_type, builder);
 
         let else_none = builder.create_block();
         builder.ins().jump(else_none, &[]);
 
-        let else_values = context.eval_all_in_block(&self.otherwise, else_none, builder);
+        let else_values = context.eval_all_in_block(&self.rhs, else_none, builder);
```
jfecher, line 176: > `builder.ins().jump(else_none, &[]);`

This is an unconditional jump to the else block, but we only want to jump there conditionally if the tag of the value returned is `None`. I think an easier alternative rather than trying to get codegen working here would be desugaring the `a else b` expressions during name resolution or type checking into:
```hs
match a
| Some value -> value
| None -> b
```

jfecher:
Since this PR has been open for some time, let me know if you want to pair program this together some time, want me to take it over entirely, or if you want to keep working on it solo. I've been rather slow to respond lately so I apologize for slowing things down a bit.

anoojpatel:
Hi! No worries, and sorry for the slow progress! I'm definitely open to pair program whenever you might be available! I'm hoping to finish this and merge it in

anoojpatel:
Thanks for the pairing session! Was a great learning experience and thanks for taking the time out :) 

To summarize, things that are left is mainly handling the single line `if a then b else c` disambiguating by creating a `without_else_expr` with a boolean value to handle that case in the parser.

jfecher:
This has gone stale and the language's design has changed to use a `Fail` ability more often where I'm unsure if this is necessary any more, so I'm closing the PR.

Issue #202, opened 2024-08-11
Remove unneeded parenthesis from type printing


jfecher:
Currently while displaying types we show an unnecessary amount of parenthesis. This can be seen while printing error messages.

E.g. `examples/typechecking/completeness_checking.an` has the following as its last error message:
```
completeness_checking.an:25:4	error: This pattern of type ((Int a), (Int b)) does not match the type ((Int a), ((Int b), ((Int c), (Int d)))) that is being matched on
| (1, 2) -> 1
```
Ideally we would display the following instead:
```
completeness_checking.an:25:4	error: This pattern of type (Int a, Int b) does not match the type (Int a, (Int b, (Int c, Int d))) that is being matched on
| (1, 2) -> 1
```
Or even the following since `,` is right-associative:
```
completeness_checking.an:25:4	error: This pattern of type Int a, Int b does not match the type Int a, Int b, Int c, Int d that is being matched on
| (1, 2) -> 1
```

The file to change for this issue would be `src/types/typeprinter.rs`. I imagine instead of always parenthesizing certain types we'll want to never do so on the type itself, and only parenthesize on certain argument positions. Either by checking the type variant or a check like `!type_string.contains_spaces()`.

jfecher:
Implemented by @eldesh in https://github.com/jfecher/ante/pull/203

PR #203, opened 2024-11-08
Remove unneeded parenthesis

eldesh:
refs #202 

Remove redundant parentheses from the type display.
No changes have been made to effect(can) as no decision was made about it.

# Algorithm.

Define the precedence of the binding of each syntactic element as follows.

[primitives > application > forall > pair > fun](https://github.com/eldesh/ante/blob/f51d34710f67fc7c14147138fe00fe70b0e559cd/src/types/mod.rs#L46-L52)

> [!NOTE] 
> The primitives includes generic Int/Float. Then the type `Foo String Int a` means `Foo String (Int a)`

In the display, brackets are added if the priority of the current object is greater than the priority of its sub-element.
The following is pseudo code.

```
fn fmt_A(&self, ...) {
  self.fmt(..)) ; // print current construct A

  if A.priority() > self.part.priority()
    write("("));
  self.fmt_part();
  if A.priority() > self.part.priority()
    write("("));

  self.fmt(..)) ; // print current construct A
}
````

# Result

For completeness_checking.rs the result is as follows:.
`(,)` is right-associative, so the parentheses are removed.

- before.

    ```console
    completeness_checking.an:25:4 error: This pattern of type ((Int a), (Int b)) does not match the type ((Int a), ((Int b)), ((Int c)), (Int d)))) that is being matched on
    | (1, 2) -> 1
    ````

- after

    ````console
    ...
    completeness_checking.an:25:4 error:. This pattern of type Int a, Int b does not match the type Int a, Int b, Int c, Int d that is being matched on
    | (1, 2) -> 1
    ````

For instantiation.rs the following applies.
`(=>)` is also right-associative, so `(a => (b => e))` becomes `a => b => e`.

- before.

    ```console
    add : (forall a b c d e f g h i. ((a - c => e can g) - (a - b => c can g) -> (a => (b => e can g) can h) can i))
    id : (forall a b. (a -> a can b))
    one : (forall a b c d. ((a => b can d) - a -> b can d))
    two1 : (forall a b c. ((b => b can c) - b -> b can c))
    two2 : ((a => a can c) => (a => a can c) can d)
    ```

- after.

    ```console
    add : forall a b c d e f g h i. ((a - c => e can g) - (a - b => c can g) -> a => b => e can g can h can i)
    id : forall a b. (a -> a can b)
    one : forall a b c d. ((a => b can d) - a -> b can d)
    two1 : forall a b c. ((b => b can c) - b -> b can c)
    two2 : (a => a can c) => a => a can c can d
    ```

### DIFF
--- a/examples/typechecking/named_constructor.an
+++ b/examples/typechecking/named_constructor.an
@@ -8,6 +8,6 @@ foo = hello_foo 42
 
 // args: --check --show-types
 // expected stdout:
-// Foo : (forall a b c. (a - b -> (Foo a b) can c))
-// foo : (Foo String (Int a))
-// hello_foo : (forall a b. (a -> (Foo String a) can b))
+// Foo : forall a b c. (a - b -> Foo a b can c)
+// foo : Foo String Int a
+// hello_foo : forall a b. (a -> Foo String a can b)
--- a/examples/typechecking/named_constructor.an
+++ b/examples/typechecking/named_constructor.an
@@ -9,5 +9,5 @@ foo = hello_foo 42
 // args: --check --show-types
 // expected stdout:
 // Foo : forall a b c. (a - b -> Foo a b can c)
-// foo : Foo String Int a
+// foo : Foo String (Int a)
 // hello_foo : forall a b. (a -> Foo String a can b)
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -45,8 +45,9 @@ impl std::fmt::Display for TypePriority {
 
 impl TypePriority {
     pub const MAX: TypePriority = TypePriority(u8::MAX);
-    pub const APP: TypePriority = TypePriority(3);
-    pub const FORALL: TypePriority = TypePriority(2);
+    pub const APP: TypePriority = TypePriority(4);
+    pub const FORALL: TypePriority = TypePriority(3);
+    pub const PAIR: TypePriority = TypePriority(2);
     pub const FUN: TypePriority = TypePriority(1);
 }
 
@@ -245,7 +246,13 @@ impl Type {
         match self {
             Primitive(_) | TypeVariable(_) | UserDefined(_) | Struct(_, _) | Tag(_) => TypePriority::MAX,
             Function(_) => TypePriority::FUN,
-            TypeApplication(_, _) => TypePriority::APP,
+            TypeApplication(ctor, _) => {
+                if ctor.is_pair_type() {
+                    TypePriority::PAIR
+                } else {
+                    TypePriority::APP
+                }
+            },
             Ref { .. } => TypePriority::APP,
             Effects(_) => unimplemented!("Type::priority for Effects"),
         }
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -250,6 +250,8 @@ impl Type {
                 TypeBinding::Unbound(..) => TypePriority::MAX,
             },
             Function(_) => TypePriority::FUN,
+            TypeApplication(ctor, _) if ctor.is_polymorphic_int_type() => TypePriority::MAX,
+            TypeApplication(ctor, _) if ctor.is_polymorphic_float_type() => TypePriority::MAX,
             TypeApplication(ctor, _) => {
                 if ctor.is_pair_type() {
                     TypePriority::PAIR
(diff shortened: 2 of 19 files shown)

### CODE REVIEW

File: examples/typechecking/named_constructor.an [hunk 1]
```diff
@@ -8,6 +8,6 @@ foo = hello_foo 42
 
 // args: --check --show-types
 // expected stdout:
-// Foo : (forall a b c. (a - b -> (Foo a b) can c))
-// foo : (Foo String (Int a))
```
jfecher, line 12: The removal of parenthesis here changes the meaning of the type since `Foo String Int a` would be a type with 3 arguments.

Can we keep the parenthesis so a type like `Foo (Int a) String` would still display correctly?

src/types/mod.rs
```diff
@@ -217,6 +241,29 @@ impl Type {
...
+        use Type::*;
+        match self {
+            Primitive(_) | UserDefined(_) | Struct(_, _) | Tag(_) => TypePriority::MAX,
+            TypeVariable(id) => match &cache.type_bindings[id.0] {
+                TypeBinding::Bound(typ) => typ.priority(cache),
+                TypeBinding::Unbound(..) => TypePriority::MAX,
+            },
+            Function(_) => TypePriority::FUN,
+            TypeApplication(ctor, _) if ctor.is_polymorphic_int_type() => TypePriority::MAX,
+            TypeApplication(ctor, _) if ctor.is_polymorphic_float_type() => TypePriority::MAX,
```
jfecher, line 254:
```suggestion
```
These cases should be removed I think. I haven't looked at it in a bit but I'm assuming these must have been included because of an existing parsing issue or similar but the intention was for `Int a` to parse like any other type application.

jfecher (commented):
This looks amazing! I only have one comment related to the precedence of `Int a` and `Float a`

src/types/mod.rs
eldesh: Yes.
These branches give special treatment to polymorphic numeric types. If these branches were simply removed, the representation of concrete types such as `I32` would always be accompanied by parentheses, as in `(I32)`.

```
// removed `TypeApplication(ctor, _) if ctor.is_polymorphic_int_type() => TypePriority::MAX,`
examples/typechecking/bind.an: Actual stdout differs from expected stdout:
  1| add_one : forall a. (Maybe I32 -> Maybe I32 can a)
   | add_one : forall a. (Maybe (I32) -> Maybe (I32) can a)
..
  4| x : Maybe I32
   | x : Maybe (I32)
```

Is there any way to properly distinguish between these?

jfecher: Oh, good question! Yes, you can distinguish between them. So when you have `Int I32` we try to display it as just `I32` but internally it will still be `TypeApplication(Type::Primitive(PrimitiveType::IntegerType), vec![i32 type])`. So you can check for that case specifically by checking what the type variable `a` is bound to in `Int a` (if it is bound).

You can find code that checks for this case in `fmt_type_application` - the `is_polymorphic_int_type` check on the constructor. Then you can see how `fmt_polymorphic_numeral` afterward checks the type bindings on the first argument to see whether it is a type variable (unbound) or already bound to a type.

eldesh: The point raised have been corrected 💪

src/types/mod.rs
```diff
@@ -250,8 +250,24 @@ impl Type {
...
+            },
+            TypeApplication(ctor, args) if ctor.is_polymorphic_float_type() => {
+                if matches!(cache.follow_typebindings_shallow(&args[0]), Type::TypeVariable(_)) {
+                    // type variable is unbound variable
+                    TypePriority::APP
+                } else {
+                    // type variable is bound (polymorphic float)
+                    TypePriority::MAX
+                }
+            },
```
jfecher, line 270:
```suggestion
            TypeApplication(ctor, args) if ctor.is_polymorphic_int_type() || ctor.is_polymorphic_float_type() => {
                if matches!(cache.follow_typebindings_shallow(&args[0]), Type::TypeVariable(_)) {
                    // type variable is unbound variable
                    TypePriority::APP
                } else {
                    // type variable is bound (polymorphic int)
                    TypePriority::MAX
                }
            },
```

jfecher (dismissed):
Looks great, thank you!

jfecher (approved):

PR #213, opened 2025-04-26
Fix #212: Unbound type from field access

### LINKED ISSUE #212: Mutable field reference of local variable has unbound type
```hs
tuple = mut ("hello", 0)

imm_ref = (&tuple).&first
imm_val = tuple.&first
mut_ref = (!tuple).!first
mut_val = tuple.!first
```

Running `ante -t ./test.an` shows:

```c
imm_ref : & bString
imm_val : & bString
mut_ref : ! bString
mut_val : ! bc
```

Here, the type of all these declarations is correct, except for `mut_val` whose type includes `c`, which is an unbound type variable.

The root cause is probably this code here:
https://github.com/jfecher/ante/blob/76ee6728188f83ca6d3ac001472f8dd80ec6fb6e/src/types/typechecker.rs#L2017-L2027

If the unification on line 2021 succeeds, then:

- `expected` is unified with `result.typ`, a mutable reference to the structure's type
- `collection_variable` is the structure's type
- the structure's type is written to `result.typ`
- everything is fine!

If the unification on line 2021 fails, on the other hand, `collection_variable` (still unbound!) is written to `result.typ`.

Later, `result.typ` is unified with `struct_type`. If `result.typ` is unbound, this leaves `field_type` unbound as well. This is then converted to a mutable-reference-to-unbound on line 2041.

(Tangent: The unification of `result.type` with `struct_type` relies on unifying a structure with a reference; this is what gives `imm_val` (see above) the correct type. It feels weird to me that `get_fields` looks through references, when no other unification does. I have no experience with any kind of type checking, so this is probably just me!)

To me, it looks like the fix is as simple as removing the assignment into `result.typ`, as that's the code path immutable references already go through. This fixes the above example, and all tests still pass.

Even though the typechecker is wrong, the backend correctly emits a field offset into the local variable. It's easy to turn this into memory unsafety:

```c
// Let's change the string's length to 16, causing an out-of-bounds read.
((tuple.!first) : !(Usz, Usz)).!second := 16
other_string = "world"
print tuple.first // `helloworld` (on my machine)

((tuple.!first) : !(Usz, Usz, I32)).!second.!second := 42
print tuple.second // `42`
```

(If you're curious: I found this by trying out `print @(mutable_local.!first)`, which reported an error when all 16 impls matched!)

LiterallyVoid:
As far as I can tell, this causes no behavior change and fixes #212.

This is a first-pass fix with very little background knowledge; let me know if some other change would be better!

### DIFF
--- a/examples/regressions/212_mutable_reference.an
+++ b/examples/regressions/212_mutable_reference.an
@@ -1,0 +1,9 @@
+a = mut (0_i32, 0_i32)
+
+field = a.!first
+field := 1_u32
+
+// args: --check
+// expected stderr:
+// 212_mutable_reference.an:4:1	error: Cannot assign expression of type I32 to a Ref of type U32
+// field := 1_u32
--- a/src/types/typechecker.rs
+++ b/src/types/typechecker.rs
@@ -2007,7 +2007,7 @@ impl<'a> Inferable<'a> for ast::Extern<'a> {
 
 impl<'a> Inferable<'a> for ast::MemberAccess<'a> {
     fn infer_impl(&mut self, cache: &mut ModuleCache<'a>) -> TypeResult {
-        let mut result = infer(self.lhs.as_mut(), cache);
+        let result = infer(self.lhs.as_mut(), cache);
 
         let level = LetBindingLevel(CURRENT_LEVEL.load(Ordering::SeqCst));
         let mut field_type = cache.next_type_variable(level);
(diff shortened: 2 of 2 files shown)


LiterallyVoid:
Actually, it seems like the code should be restructured so that `get_field` doesn't look through references. As is (and unaffected by this fix), the following typechecks:

```c
print (&&(1, 1)).first
```

and immediately panics during monomorphisation.

Furthermore, this can be used to launder mutability from immutability:

```c
immutable_tuple = (1, 1)
mut_reference = (!&immutable_tuple).!first
mut_reference := 3
```

This typechecks! But it still panics during monomorphisation, so there's no exploit here :)

LiterallyVoid:
Obsoleted by merge of #214

jfecher:
Thanks for the investigation @LiterallyVoid !

PR #214, opened 2025-04-26
Move deref typecheck from `get_fields` to `MemberAccess`

LiterallyVoid:
I was curious how hard this would be (see https://github.com/jfecher/ante/pull/213#issuecomment-2832021096), and it wasn't actually *that* bad!

This now only allows a single dereference on field access, where previously, multiple dereferences passed typechecking (although this failed later in compilation).

Feel free to not merge this. I don't know what behaviour is desired, and this patch is a little ugly :)

This fixes a typechecking safety hole:

```c
immutable_tuple = (1, 1)
mut_reference = (!&immutable_tuple).!first
mut_reference := 3
```

which now fails with this error:
```
test.an:4:6     error: !shared &shared T has no field 'a' of type { a: b, .. }
a = (!&tuple).!a
```

### DIFF
--- a/examples/typechecking/field_diagnostics.an
+++ b/examples/typechecking/field_diagnostics.an
@@ -1,0 +1,54 @@
+tuple = (1, 1)
+
+_ = (&&tuple).first
+
+_ = tuple.!first
+_ = tuple.!missing
+
+imm_ref = &tuple
+_ = imm_ref.!first
+_ = imm_ref.!missing
+
+mut_tuple = mut (1, 1)
+
+_ = mut_tuple.&missing
+_ = mut_tuple.!missing
+
+mut_ref = !mut_tuple
+
+_ = mut_ref.&missing
+_ = mut_ref.!missing
+
+_ = (!&tuple).!first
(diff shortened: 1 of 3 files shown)


jfecher (approved):
Had to double-check the approach here. It looks like the issue is that `get_fields` originally retrieved fields through references, which you removed, but since `get_fields` was used in `try_unify_with_bindings_inner` when type checking a struct type against another arbitrary type it could lead to us accidentally unifying a `Struct` with a `&Struct`, so I think this fix is the correct one - thank you!

jfecher:
Failing CI is due to it not being updated to pull submodules - now fixed on master.

Verified locally this is compiling and passing all tests - merging now.

jfecher:
> which now fails with this error:

```
test.an:4:6     error: !shared &shared T has no field 'a' of type { a: b, .. }
a = (!&tuple).!a
```
Ante's expected field type errors can definitely be improved...
Thanks for the contribution! I appreciate your digging into the type system which can be intimidating at first

PR #220, opened 2025-10-28
Incremental: Named Constructor Parsing

Fmurph97:


src/parser/mod.rs
```diff
@@ -1200,10 +1200,63 @@ impl<'tokens> Parser<'tokens> {
...
+        while let Ok(field) = self.parse_named_constructor_field() {
+            fields.push(field);
+
+            if !self.accept(Token::Comma) {
+                break;
+            }
+        }
+        
+        Ok(fields)
+    }
```
jfecher: function can be replaced with `this.delimited(Self::parse_named_constructor_field, Token::Comma, true);`

jfecher (changes requested):

PR #225, opened 2026-04-30
Have the CLI properly error with no source files

hatzka-nezumi:
Without this it was successfully "compiling" the empty program and then failing to run it because it has no main function.


jfecher (approved):
LGTM

CI Error just shows a test with an old flag erroring the flag no longer exists, thus showing the changed cli format. I'll fix it in another commit.

PR #227, opened 2026-05-02
Refactor flake.nix and friends

polygonalbones:
Fixed flake.nix, now you can build the latest version of Ante. Also removed a lot of legacy Nix compatibility (default.nix, shell.nix) and merged it into the flake.

This should also (hopefully) make it easier for other people to maintain any nix-related things.


jfecher (dismissed):
Amazing, thank you!

I'll be honest - I don't have a good way to test this - but it looks good to me. I'll double-check with you though: have you verified the resulting binary compiles & that it can compile & run the examples in the `examples` directory? (you can run `cargo t` to check)

jfecher (approved):
Thanks!

PR #228, opened 2026-05-02
Fix nix package

polygonalbones:
Fixes the Nix package breakage. Really sorry, I forgot last night before pushing to remove the commented code and `nix run` was using the cached version of Ante, so the bug wasn't triggering 😭.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -56,8 +56,6 @@
                         ./Cargo.toml
                         ./Cargo.lock
                         ./build.rs
-                        ./tests
-                        ./goldentests.toml
                       ];
                     };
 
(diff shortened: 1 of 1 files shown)


Shakadak:
Thanks, now it does build, but running it on a file fails:
```
➜  typechecking git:(master) nix run github:polygonalbones/ante -- --check affine.an

thread 'main' (36018) panicked at src/paths.rs:28:23:
Failed to canonicalize stdlib path /build/source/stdlib ; does it exist?
```

polygonalbones:
dw I'll get to fixing it

polygonalbones:
I'm gonna have to close for now, I think adding `ANTE_STDLIB_DIR = "${ante-stdlib}/lib";` to the `ante` package would fix it, but unfortunately I can't test it for now because my bad RAM caught up to me and Ante just refuses to compile now, and I don't wanna PR without testing it on my end 😭😭

PR #231, opened 2026-05-13
Add a hint to errors about fieldless types (and put parser errors on the right line)

hatzka-nezumi:
This adds a hint to syntax errors that look like my attempt to make a fieldless type. I also found and fixed a semi-bug where if an unexpected newline was found, the error would be reported on the next line.

For some reason I wasn't able to make the hint work if there's an `=` followed by nothing, so I left a todo there.

### DIFF
--- a/src/diagnostics/mod.rs
+++ b/src/diagnostics/mod.rs
@@ -22,8 +22,6 @@ pub use location::*;
 pub use unimplemented_item::*;
 
 /// Any diagnostic that the compiler can issue.
-///
-/// If the `hint` is `None`, it shouldn't be printed.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
 pub enum Diagnostic {
     // TODO: `message` could be an enum to save allocation costs
--- a/src/diagnostics/mod.rs
+++ b/src/diagnostics/mod.rs
@@ -1,6 +1,5 @@
-use std::{cmp::Ordering, collections::BTreeSet, fmt::Formatter, path::PathBuf, sync::Arc};
+use std::{cmp::Ordering, collections::BTreeSet, fmt::{Display, Formatter}, path::PathBuf, sync::Arc};
 
-use beef::lean::Cow;
 use colored::{Color, ColoredString, Colorize};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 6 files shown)


src/diagnostics/mod.rs
```diff
@@ -313,13 +349,32 @@ impl Diagnostic {
...
     pub fn message(&self) -> String {
         match self {
-            Diagnostic::ParserExpected { message, actual, location: _ } => {
+            Diagnostic::ParserExpected { message, actual, location: _, hint } => {
                 if actual.to_string().contains(" ") {
-                    format!("Expected {message} but found {actual}")
+                    if hint.is_empty() {
+                        format!("Expected {message} but found {actual}")
+                    } else {
+                        format!("Expected {message} but found {actual} ({hint})")
```
jfecher: I think it'd make more sense to put the hint in the note, see `fn note` further below

src/diagnostics/mod.rs
```diff
@@ -21,14 +22,49 @@ mod unimplemented_item;
...
+/// If the `hint` is the empty string, it shouldn't be printed.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Hash)]
 pub enum Diagnostic {
     // TODO: `message` could be an enum to save allocation costs
     ParserExpected {
         message: String,
         actual: Token,
         location: Location,
+        #[serde(skip_serializing_if = "str::is_empty", default, with = "serde_beef")]
+        hint: Cow<'static, str>,
```
jfecher: No need to have this be a Cow, it is only ever a `&'static str`. Let's do that and remove the serde_beef dependency.

src/parser/mod.rs
```diff
@@ -765,6 +773,7 @@ impl<'tokens> Parser<'tokens> {
             },
             _ => match self.parse_type() {
                 Ok(typ) => Ok(TypeDefinitionBody::Alias(typ)),
+                // TODO: hint about fieldless types here if appropriate
```
jfecher:
```suggestion
```
It shouldn't be possible to have no other tokens within an `<indent> <unindent>` pair

jfecher (commented):
Thanks for the PR and test! I do have some small suggestions before merging

src/diagnostics/mod.rs
hatzka-nezumi: It's only ever explicitly set to a `&'static str`, but it's a `String` if it came from deserialization. I could make it an enum, but it seems a bit boilerplatey.

src/parser/mod.rs
hatzka-nezumi: This is where I tried to insert a hint for the version of the error you get with `type Bad =` and nothing else. If it's in the wrong place, that's probably why it didn't work...

src/diagnostics/mod.rs
jfecher: Do you mind making it an optional enum? I'd mainly like to avoid the extra dependency for now

hatzka-nezumi: Sure.

src/diagnostics/mod.rs
```diff
@@ -21,14 +21,17 @@ mod unimplemented_item;
 pub use location::*;
 pub use unimplemented_item::*;
 
-/// Any diagnostic that the compiler can issue
+/// Any diagnostic that the compiler can issue.
+///
+/// If the `hint` is `None`, it shouldn't be printed.
```
jfecher:
```suggestion
```
`hint` is only on the ParserExpected variant so we don't need to document it on the enum itself

jfecher (approved):
Thanks for the PR and for the requested changes!

Ante's parser needs a lot of work making its errors more friendly so this is one step towards that!

Issue #235, opened 2026-05-29
Add a subcommand interface for CLI


jfecher:
- [ ] `ante file.an`: Compile a single ante file, similar to today but do not run it afterward.
  - This works but needs to be changed to just compile instead of compile & run by default
- [x] `ante build`: Build a project. There's no project file for ante yet so it is unclear which file we should start with. It'll likely end up looking for main.an for applications or lib.an for libraries.
- [x] `ante run`: Build & run a project, forwarding any CLI args.
- [x] `ante add <dep-url>`: Adds the git dependency at the given repository URL as a dependency of the local project. This is just a git clone into the deps directory. In the future dependencies will be de-duped across all projects on a system.

New:
- [x] Integrate existing compiler options like `--backend c` so they can be used with `ante build` and `ante run` subcommands where appropriate. We should go through each option and divide them into a `CompileArgs` struct of options that make sense for any compilation command and include that struct in the arguments for e.g. `build`, `run`, and the default when no subcommand is specified.

mprakriti04-glitch:
Hi! I'm new to open source contributions and I'm interested in working on this issue if it's available. I'd love to give it a try and learn from the process.

jfecher:
Hi @mprakriti04-glitch that'd be great! Feel free to fork & create a PR here when you're ready.

This change will probably end up touching `src/main.rs` and `src/cli.rs`. The later contains the `Cli` struct where you can add new commands via the `clap` library. For the initial PR feel free to start it with just one command if you get stuck and I can help you through the rest. How PRs usually go is once they're made I'll suggest a few changes for you to make, and once everything is resolved it can be merged.

mprakriti04-glitch:
Thanks for the guidance! I'll fork the repo and start exploring the CLI code.

mprakriti04-glitch:
I have been reading through "cli.rs" and "main.rs" and I think I'm starting to understand how the current flag-based CLI works. For my first PR I'm planning to focus only on implementing "ante build"

My current understanding of the issue is that "ante file.an" should continue to work, while adding subcommands like "ante build", "ante run" etc. Is that the intended direction, or are you aiming to fully switch to a subcommand-based interface?

I'm still pretty new to open source and to navigating larger codebases, so I wanted to make sure I understood the goal correctly before I start making changes.

jfecher:
[...]

PR #243, opened 2026-07-09
Add `ante add` command for git dependencies

rupert648:
Hello! I thought I'd have a crack at tackling https://github.com/jfecher/ante/issues/235 . `ante add` specifically
This PR implements the most naive approach - that is simply cloning the dependency via running Git.

Theres certain more we could iterate on here with dependency handling via `ante.toml`. Further "sanity" checking of dependencies at "ante add" time instead of build time (e.g. requiring ante.toml file in dependencies - or even all ante projects?). And of course eventually going the full monty and adding a registry/versioned dependencies/checksums etc. But since this is my first contribution I wanted to keep the scope small and useful, but I'd love to iterate on some of these things together.

Theres also a second commit - I fixed ante run for projects without a manifest name: the compiler was deriving the fallback output path inconsistently (a.out became a) and then trying to spawn the bare filename. It now uses a consistently and runs local binaries via ./a.

# Testing

For testing I made the classic is-even dependency - https://github.com/rupert648/ante-is-even
And ran through the workflow of adding it to a mini CLI project (P.S. am I being dumb and missing a parse_i32 in stdlib?)
```
import Std.Env
import Std.Vec
import Std.Stream.stream_fn
import Std.Char.is_digit
import IsEven.Lib.is_even

parse_i32 (s: String): Maybe I32 =
    if s.length == 0 then return Maybe.None

    loop (i = 0usz) (value = 0) ->
        if i == Usz s.length then return Maybe.Some value

        c = s.data.[i]
        if not is_digit c then return Maybe.None

        recur (i + 1) (value * 10 + (I32 c - I32 '0'))

main () =
    args = Vec.of Env.args
    if args.len != 2 then
        println "usage: ante-add-test <number>"
        return ()

    match parse_i32 args.data.[1]
    | None -> println "error: expected a non-negative integer"
    | Some n ->
        if is_even n then
            println ((cast n: String) ++ " is even")
        else
            println ((cast n: String) ++ " is odd")
```

Then running/adding dependency
```sh
~/Documents/personal/ante-add-test ⌚ 17:16:46
$ /Users/rupert/Documents/personal/ante/target/debug/ante build
/Users/rupert/Documents/personal/ante-add-test/src/main.an:5:8
4 | import Std.Char.is_digit
5 | import IsEven.Lib.is_even   // error: Could not find module `Lib` in crate `IsEven`

/Users/rupert/Documents/personal/ante-add-test/src/main.an:27:12
26 |     | Some n ->
27 |         if is_even n then  // error: `is_even` not found in scope
28 |             println ((cast n: String) ++ " is even")

Found 2 errors and 0 warnings

~/Documents/personal/ante-add-test ⌚ 17:16:48
$ /Users/rupert/Documents/personal/ante/target/debug/ante add git@github.com:rupert648/ante-is-even.git
Cloning into 'ante-is-even'...
remote: Enumerating objects: 8, done.
remote: Counting objects: 100% (8/8), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 8 (delta 0), reused 8 (delta 0), pack-reused 0 (from 0)
Receiving objects: 100% (8/8), done.

~/Documents/personal/ante-add-test ⌚ 17:16:55
$ /Users/rupert/Documents/personal/ante/target/debug/ante build

~/Documents/personal/ante-add-test ⌚ 17:17:03
$ ./a hello
error: expected a non-negative integer

~/Documents/personal/ante-add-test ⌚ 17:17:11
$ ./a 42
42 is even

~/
```

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -12,6 +12,12 @@ pub struct Completions {
 pub enum Commands {
     Build,
     Run,
+    /// Clone a git dependency into this project's dependency directory
+    Add {
+        /// Git repository URL to add as a dependency
+        #[arg(value_hint = ValueHint::Url)]
+        dep_url: String,
+    },
 }
 #[derive(Parser, Debug)]
 #[command(author, version, about, long_about = None)]
--- a/src/dependencies.rs
+++ b/src/dependencies.rs
@@ -1,0 +1,38 @@
+use std::process::Command;
+
+use colored::Colorize;
+
+use crate::{find_files::find_nearest_project_root, manifest::MANIFEST_FILE_NAME};
+
+pub const DEPENDENCIES_DIR_NAME: &str = "deps";
+
+/// Add a git dependency to the current Ante project by cloning it into the dependencies directory.
+pub fn add_git_dependency(dep_url: &str) {
+    let Some(project_root) = find_nearest_project_root() else {
+        eprintln!(
+            "{}: could not find {MANIFEST_FILE_NAME} in the current directory or any parent directory",
+            "error".red(),
+        );
+        std::process::exit(1);
+    };
+
+    let deps_dir = project_root.join(DEPENDENCIES_DIR_NAME);
+    if let Err(error) = std::fs::create_dir_all(&deps_dir) {
+        eprintln!("{}: failed to create `{}`:\n{error}", "error".red(), deps_dir.display());
+        std::process::exit(1);
+    }
+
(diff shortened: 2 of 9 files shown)


jfecher:
> am I being dumb and missing a parse_i32 in stdlib?

It used to have `parse_int` but I removed it when doing the compiler rewrite and I suppose it was never re-added. If you want to take a stab at finding & adding missing stdlib functions / types, that'd also be appreciated!

jfecher (approved):
LGTM! Thank you for the PR and the directory constants refactor!

PR #247, opened 2026-07-12
Add integer overflow intrinsics and first i32 math helpers

rupert648:
Following on from this comment https://github.com/jfecher/ante/pull/243#issuecomment-4927724480 . I started implementing a parse_int function. This was starting with a `parse_i32` function. After doing some research on the [rust implementation](https://doc.rust-lang.org/src/core/num/mod.rs.html#1756) I noticed that it used checked_* implementations. (It also has a fast/happy path where it does some basic string heuristics to see if it can skip the checked path: https://doc.rust-lang.org/src/core/num/mod.rs.html#1569 - we could add these in the future). This then led me down the rabbit hole of intrinsics in ante - and I noticed that these weren't present for add|sub|mul_overflow. Adding these is relatively simple and is supported by both llvm & c.

These return a bool as a secondary return value, which in the stdlib we simply change to a nicer signature like so:
```ante
checked_add_i32 (x: I32) (y: I32): Maybe I32 =
    result, overflowed = overflowing_add_i32 x y
    if overflowed then None else Some result
```
where overflowing_add_i32 looks like so:
```
overflowing_add_i32 (x: I32) (y: I32): (I32, Bool) =
    intrinsic "OverflowingAddInt" x y
```
We have the same setup for checked_sub/mul.

Following PRs (see list at bottom) then expand these to cover other integer types and try and make some generic-ness to re-use some of the code using ante's Abilities. These are the main ones where I'd love your opinion on the approach - this MR simply sets up how intrinsic usage will look like.

We also later use these for string parsing of integers, which matches what rust does. I also, in later MRs introduce a FromStr ability (stolen from rust). We then implement these for all basic integer types. This will allow the following:
```
    from_u64: Maybe U64 = from_str "42"
    println from_u64
    from_u64_overflow: Maybe U64 = from_str "18446744073709551616"
    println from_u64_overflow
    from_usz: Maybe Usz = from_str "12345"
```

The subsequent PRs, these have to exist on my fork as they can't target my branches otherwise. I'll open subsequent PRs as these are merged.
- https://github.com/rupert648/ante/pull/1
- https://github.com/rupert648/ante/pull/2
- https://github.com/rupert648/ante/pull/3


jfecher (approved):
Absolutely amazing work in this PR! 😄

- I'm not 100% sure on the organization I'd want on these stdlib items (e.g. if they should be renamed) but they can always be moved later.

- I'm also debating whether `checked_*` should use `Fail` effects instead of returning `Maybe` values. I'll likely change this soon. The stdlib is a bit inconsistent here as well and there's a real tradeoff between them. E.g. `println (checked_add_i32 ...)` would stop working since you'd need to handle the fail effect, differences in efficiency, etc.

- Unfortunately due to C it seems Ante will need a bunch of intrinsics for checked operations for each integer size. Future work would be adding these for more integer sizes and seeing if there is a cleaner way to handle multiple.

> I also, in later MRs introduce a FromStr ability (stolen from rust)

I've yet to look into the other changes yet (I'll leave more comments after this review for them), but I'd like to avoid a `FromStr` ability in favor of the existing `Cast` and `TryCast` abilities. `FromStr` is just a specialized `TryCast` essentially so I'd like to avoid the duplicates.

jfecher:
For the other branches:
- https://github.com/rupert648/ante/pull/1: I'd prefer a `Fail` effect on parse_i32 over returning None. I just pushed a `fail_if` helper function to the stdlib which may be helpful.
- https://github.com/rupert648/ante/pull/2: Good, but I'd prefer a `Fail` effect for the checked methods again, and to use `TryCast` over `FromStr`.
- https://github.com/rupert648/ante/pull/3: no notes 👍 . Since you kept the 1 `OverflowingAddInt` intrinsic, it looks like I was wrong about C only supporting that for `int`? I see e.g. checked_mul_i64 being used without returning None so I suppose I was wrong. All these printouts being in a separate location is getting somewhat unwieldy, we may want to switch to asserts but that is something I can change later.

PR #248, opened 2026-07-13
Working/lexer new escape sequences

### LINKED ISSUE #245: Add more string and char escape sequences
String escapes are handled in `src/lexer/mod.rs:336-352` and char escapes in `src/lexer/mod.rs:404-425`.

Currently only:
- `\\ \" \$ \n \r \t \0` (strings) and 
- `\\ \' \n \r \t \0` (chars) are supported

Add some more escapes:
- `\xNN` — a hex byte escape.
- `\u{...}` — a Unicode codepoint escape.

### Important

Chars in Ante are currently only 1 byte, though this may be changed in the future. This means they can't fit a full unicode code point, but we should still support the syntax and give a warning for now.

### Testing

Add these escapes to a test in examples/parser and ensure all tests pass

Stiven-Gjekaj:
##  Add `\xNN` and `\u{...}` escape sequences

Closes #245.

Adds two escape sequences to string and character literals:

- `\xNN`  a hexadecimal byte escape (exactly 2 hex digits)
- `\u{...}`  a Unicode codepoint escape (1–6 hex digits)

```ante
"\x41\x42\x43"                    // "ABC"
"\u{48}\u{65}\u{6c}\u{6c}\u{6f}"  // "Hello"
"\u{1F600}"                       // "😀"
'\x41'                            // 'A'

### CODE REVIEW

File: examples/codegen/stdlib-tests/escape_sequences.an [hunk 1]
```diff
@@ -0,0 +1,18 @@
+import Std.String.bytes
+import Std.Stream.iter, stream_fn, Emit, emit
+
+main () =
```
jfecher: This file can be deleted, the parser test is sufficient

src/lexer/mod.rs
```diff
@@ -377,35 +386,136 @@ impl<'contents> Lexer<'contents> {
...
+            'x' => {
+                self.advance();
+                self.lex_hex_escape(escape_start)
+            },
+            'u' => {
+                self.advance();
+                self.lex_unicode_escape(escape_start)
+            },
+            _ => {
+                self.advance();
```
jfecher: advance is called in every branch. We should just put it once before the match.

src/lexer/mod.rs
```diff
@@ -377,35 +386,136 @@ impl<'contents> Lexer<'contents> {
...
+            }
         }
         self.expect('"', Token::StringLiteral(contents))
     }
 
+    /// Lex a single escape sequence within a string literal, e.g. `\n` or `\x41`.
+    /// Expects `self.current == '\\'`. Always leaves `self.current` positioned just past
+    /// everything consumed as part of the escape (valid or invalid) - the caller performs
+    /// no additional advance. Returns `None` (having already recorded a [LexerError]) if
+    /// the escape was invalid, in which case nothing should be pushed to the string.
```
jfecher: Please reword this AI-generated comment

src/lexer/mod.rs
```diff
@@ -377,35 +386,136 @@ impl<'contents> Lexer<'contents> {
...
+                self.errors.push((LexerError::InvalidEscapeSequence(selector), span));
+                None
+            },
+        }
+    }
+
+    /// Lex the `NN` in a `\xNN` escape (shared by string and char literals). Expects
+    /// `self.current` to be the first hex digit (`\` and `x` are already consumed).
+    /// Consumes at most 2 hex-digit characters and never consumes a non-hex-digit, so
+    /// callers can always resume normal lexing immediately after this returns.
```
jfecher:
```suggestion
    /// Lex the `NN` in a `\xNN` escape. Expects `self.current` to be the first hex digit
```
Claude never knows when to stop talking

src/lexer/mod.rs
```diff
@@ -377,35 +386,136 @@ impl<'contents> Lexer<'contents> {
...
+            None
+        }
+    }
+
+    /// Lex the `{XXXXXX}` in a `\u{...}` escape (shared by string and char literals).
+    /// Expects `self.current` to be positioned right after `u`. A `}` immediately
+    /// following the digit scan is always consumed, even on error, since it
+    /// unambiguously belongs to this escape; a `}` that never appears (e.g. the digit
+    /// scan runs into the literal's own closing quote) is never consumed, so callers can
+    /// recover cleanly without cascading errors.
```
jfecher:
```suggestion
    /// Lex the `{XXXXXX}` in a `\u{...}` escape. Expects `self.current` to be positioned right after `u`
```

src/lexer/mod.rs
```diff
@@ -451,27 +561,74 @@ impl<'contents> Lexer<'contents> {
...
 
-        self.advance();
         self.expect('\'', Token::CharLiteral(contents))
     }
 
+    /// Lex a single escape sequence within a char literal, e.g. `\n` or `\x41`. Expects
+    /// `self.current == '\\'`. Always leaves `self.current` positioned just past
+    /// everything consumed as part of the escape - the caller performs no additional
+    /// advance. Returns `'\0'` as a placeholder if the escape was invalid, having
+    /// already recorded a [LexerError].
```
jfecher:
```suggestion
    /// Lex a single escape sequence within a char literal, e.g. `\n` or `\x41`. Expects
    /// `self.current == '\\'`.
```

src/lexer/mod.rs
```diff
@@ -451,27 +561,74 @@ impl<'contents> Lexer<'contents> {
...
+    fn lex_char_escape(&mut self) -> char {
+        let escape_start = self.current_position;
+        self.advance(); // consume '\'
+        let selector = self.current;
+
+        match selector {
+            '\\' | '\'' => {
+                self.advance();
```
jfecher: Same here with the advance

src/lexer/mod.rs
```diff
@@ -682,7 +839,10 @@ impl<'contents> Lexer<'contents> {
 
     pub fn lex(file_contents: &'contents str, id: SourceFileId) -> (Vec<(Token, Span)>, Vec<Diagnostic>) {
         let mut lexer = Lexer::new(file_contents);
-        ((&mut lexer).collect(), lexer.errors(id))
+        let tokens = (&mut lexer).collect();
+        let mut diagnostics = lexer.errors(id);
+        diagnostics.extend(lexer.warnings(id));
```
jfecher: We could avoid allocating an extra Vec if `lexer.warnings` returned an iterator instead of a Vec

src/lexer/token.rs
```diff
@@ -420,6 +429,28 @@ impl Display for LexerError {
     }
 }
 
+/// Non-fatal lexing issues. Unlike [LexerError] these still produce a normal, usable
+/// token; the token's value should be used as-is, just with the caveat this describes.
```
jfecher:
```suggestion
/// Non-fatal lexing diagnostics
```

src/lexer/token.rs
```diff
@@ -420,6 +429,28 @@ impl Display for LexerError {
...
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize, Hash)]
+pub enum LexerWarning {
+    /// A `\u{...}` escape in a char literal produced a codepoint above U+FF. Ante's
+    /// `Char` is a single byte, so this value is truncated to its low byte at codegen.
```
jfecher:
```suggestion
```

src/lexer/token.rs
```diff
@@ -420,6 +429,28 @@ impl Display for LexerError {
...
+    CodepointDoesNotFitInChar { codepoint: u32 },
+}
+
+impl Display for LexerWarning {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use LexerWarning::*;
+        match self {
+            CodepointDoesNotFitInChar { codepoint } => write!(
+                f,
+                "Codepoint U+{:x} does not fit in a single byte; this Char literal's value will be truncated to its low byte at compile time",
```
jfecher:
```suggestion
                "Codepoint U+{:x} does not fit in a single byte and will be truncated",
```

jfecher (commented):
This PR looks entirely AI-generated to me. I appreciate your interest in the language but it would show more interest if you had attempted this small change yourself or reviewed the AI output.

That being said, this is a minor change so once a couple minor comments I have are resolved we can still get this merged.

src/lexer/mod.rs
jfecher: If `errors` were changed too, they could be chained together and we could pre-reserve the whole resulting Vec's size at once since it'd be a single ExactSizeIterator.

Stiven-Gjekaj:
Thanks for the review. I've pushed a commit addressing it:

- **`advance` in every branch:** pulled it out to a single call before the match in
  both escape helpers. You're right that every arm consumed exactly the selector char,
  so it's a clean hoist. I re-checked the malformed cases (`\xG`, `\u{}`, `\u{110000}`,
  etc.) still recover the same way afterward.
- **`errors`/`warnings` allocating extra Vecs:** both now return iterators and get
  `chain`ed in `lex`, so the diagnostics Vec is allocated once with the full size
  reserved instead of building two Vecs and extending.
- **Comments + docstrings:** trimmed.
- **Codegen test:** removed it.

Full test suite passes.
Thank you for your time.

jfecher (approved):

PR #249, opened 2026-07-13
Use Fail ability for checked_i32 Math and add basic i32 string parsing

rupert648:
Switches the `checked_*_i32` functions to use `{Fail}` and updates call points. Note for now I just switched the golden callsites to do something like
```
println (try (fn f -> checked_add_i32 2_147_483_647 1 {f}) : Maybe I32)
```
so then we can use Maybe and get the same output, but we could prob add a stdlib.test assertion helper here (which just does the same thing).

The second commit adds string parsing for i32s and some tests. I've updated the [followup](https://github.com/rupert648/ante/pull/2) to implement this for all integer types and implement `TryCast` as we discussed.

### DIFF
--- a/examples/codegen/stdlib-tests/math.an
+++ b/examples/codegen/stdlib-tests/math.an
@@ -1,6 +1,7 @@
 import Std.Math.sqrt, pow, floor, ceil, abs
 import Std.Math.overflowing_add_i32, overflowing_sub_i32, overflowing_mul_i32
 import Std.Math.checked_add_i32, checked_sub_i32, checked_mul_i32
+import Std.Fail.try
 
 main () =
     println <| sqrt 16.0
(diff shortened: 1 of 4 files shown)


examples/codegen/stdlib-tests/math.an
```diff
@@ -10,13 +11,13 @@ main () =
...
     println <| abs 5
     println <| abs (0 - 12345)
-    println <| checked_add_i32 2_147_483_647 1
-    println <| checked_add_i32 40 2
-    println <| checked_sub_i32 -2_147_483_648 1
-    println <| checked_sub_i32 40 2
-    println <| checked_mul_i32 50_000 50_000
-    println <| checked_mul_i32 -2_147_483_648 -1
-    println <| checked_mul_i32 -12 3
+    println (try (fn f -> checked_add_i32 2_147_483_647 1 {f}) : Maybe I32)
```
jfecher: Do you get an error without the type annotation?

These should be able to be simplified to:

```suggestion
    println (try do checked_add_i32 2_147_483_647 1)
```

examples/codegen/stdlib-tests/string.an
```diff
@@ -48,6 +50,39 @@ test_predicates () =
     println ("[" ++ trim "   " ++ "]")
     println ("[" ++ trim "no-trim" ++ "]")
 
+try_parse_i32 (s: String): Maybe I32 =
+    try (fn f -> parse_i32 s {f})
```
jfecher: Same here w/ `try do`, then the helper wouldn't be needed any more.

You could also introduce `f` as an implicit via `try (fn {f} -> parse_i32 s)` so the compiler knows it can thread it through automatically, but the `do` is shorter in this case.

stdlib/src/String.an
```diff
@@ -139,3 +141,40 @@ trim_end (s: String): String =
...
+        i += 1
+
+    fail_if (i == len)
+
+    var result = 0i32
+
+    while i < len do
+        digit = match to_digit s.data.[i]
+        | Some d -> I32 d
+        | None -> fail ()
```
jfecher, line 169: Unrelated: I should add a `Maybe.unwrap` variant which fails instead of panics. I wonder what naming scheme the two versions of the function should use.

examples/codegen/stdlib-tests/math.an
jfecher: Going to apply this suggestion now to ensure it still compiles

stdlib/src/String.an
rupert648: perhaps `Maybe.unwrap_or_fail` if you want to keep some familiarity with rust?

jfecher (approved):
LGTM!

Thank you once again

PR #250, opened 2026-07-13
Switch TryCast to use Fail ability instead of Maybe

rupert648:
In places which used this (only Prelude.an), i just matched the behvaiour of whatever is there before
e.g.
Maybe.unwrap y-> `try_cast y -> panic_on_fail`


jfecher (approved):
Thank you!

PR #251, opened 2026-07-13
Add TryCast impls for integer string parsing

rupert648:
The big one! Most of this is trying to include relatively extensive testing (fuzz testing would be cool here!).

And then its lots of repetitions between the various integer types. Most functions being the exact same other than signature - it would be a nice place for macro/comptime to help (rust does it with a macro https://doc.rust-lang.org/src/core/num/mod.rs.html#1849).

One thing to note - I've chosen *not* to export the `overflowing_*` functions - these are the ones that look like the below and 
```
overflowing_sub_usz (x: Usz) (y: Usz): (Usz, Bool) =
    intrinsic "OverflowingSubInt" x y
```
And only export the `checked_*` functions with use `{Fail}`. I think this is the better API so we shouldnt expose those internal versions - but I think worth keeping as it prevents dotting `Intrinsic` around the place if they're used elsewhere. (this also means removing `overflowing_*_i32` functions from the last MR.

For other things to note I've added a `CheckedNum` ability
```
ability CheckedNum t =
    checked_add: fn t t {Fail} -> t
    checked_sub: fn t t {Fail} -> t
    checked_mul: fn t t {Fail} -> t
```
You'll see this makes the implementation neater by allowing us to use a generic `parse_signed_int` and `parse_unsigned_int` functions rather than having to have specifics for both. `CheckedNum` could go in Prelude.an, lmk what you think!

### DIFF
--- a/stdlib/src/String.an
+++ b/stdlib/src/String.an
@@ -3,7 +3,9 @@ import Std.C.malloc, memcpy, memcmp
 import Std.Stream.Emit, emit, Stream, foldl, map, stream_fn
 import Std.Vec.Vec
 import Std.Char.is_whitespace, to_digit
-import Std.Math.checked_add_i32, checked_sub_i32, checked_mul_i32
+import Std.Math.CheckedNum, checked_add, checked_sub, checked_mul
+import Std.Math.checked_num_i8, checked_num_i16, checked_num_i32, checked_num_i64, checked_num_isz
+import Std.Math.checked_num_u8, checked_num_u16, checked_num_u32, checked_num_u64, checked_num_usz
 import Std.Fail.Fail, fail, fail_if
 
 // TODO: All functions in this file should be a method on String, but String isn't defined in this file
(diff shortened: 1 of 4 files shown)


stdlib/src/String.an
```diff
@@ -161,20 +193,80 @@ parse_i32 (s: String) {Fail}: I32 =
...
 
-    var result = 0i32
+    ten: t = cast 10
+    var result = n.zero
 
     while i < len do
-        digit = match to_digit s.data.[i]
-        | Some d -> I32 d
+        digit: t = match to_digit s.data.[i]
+        | Some d -> cast (I32 d)
```
jfecher, line 201: This may need attention in the future but is good for now.

`cast` is meant to be an infallible cast, which would exclude truncating conversions like I32 to I8. So casting `d` to `I8` here first may be more appropriate, but this works for now while we still have the truncate impls. I'm assuming you did it this way because there's already a `Char` to `I32` cast, which makes sense.

jfecher (approved):
Amazing work again, thank you for all these PRs 😄
They've been a big help

PR #252, opened 2026-07-14
Fix stack-buffer over-read in C backend's transmute codegen

lichtblau:
Fix stack-buffer over-read in C backend's transmute codegen

Unlike the LLVM bitcast, the C backend transmutes using memcpy. If destination > source, the source would be over-read.  Clamp it.

This upstreams commit af40ebcee1776b5d8dc81f0574d7ff31783599c9 from the "drop" feature branch.


jfecher (approved):
Thank you

PR #253, opened 2026-07-15
Enable -fno-strict-aliasing for the C backend

lichtblau:
Enable -fno-strict-aliasing for the C backend

### DIFF
--- a/examples/codegen/strict_aliasing.an
+++ b/examples/codegen/strict_aliasing.an
@@ -1,0 +1,16 @@
+// Caveat: This would cease to be a good test if the compiler were to inline the helper function
+// and constant-fold it.  4602678819172646912 = 0x3FE0000000000000
+
+bits_roundtrip (i: mut I64) (f: mut F64): I64 =
+    i.* := 11
+    f.* := 0.5
+    i.*
+
+main () =
+    var x = 0 : I64
+    f = transmute (mut x) : mut F64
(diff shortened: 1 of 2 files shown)


examples/codegen/strict_aliasing.an
```diff
@@ -0,0 +1,16 @@
+// Caveat: This would cease to be a good test if the compiler were to inline the helper function
+// and constant-fold it.  4602678819172646912 = 0x3FE0000000000000
+
+bits_roundtrip (i: mut I64) (f: mut F64): I64 =
```
lichtblau, line 4: This change is being proposed based on a real code base failing otherwise.

The test case offered to demonstrate the issue is a bit on the contrived side admittedly...

jfecher (approved):
We can merge this to be on the safer side, though I'm curious what the real world code you were using that ran into this looked like. Was it casting union variants?

lichtblau:
> We can merge this to be on the safer side, though I'm curious what the real world code you were using that ran into this looked like. Was it casting union variants?

I wish I had a more exciting reproduction but I couldn't come up with it.  For all I know, I'm just calling cranelift through FFI, but most of the program was just "optimized" away entirely.  (I had originally expected the issue in some GC code in the same program, where there are plenty of raw pointers, but somehow that's not the part that first triggered it.)  Apologies that I don't have anything more specific to share.

PR #254, opened 2026-07-20
Patch minicoro for cross-coroutine longjmp

lichtblau:
Patch minicoro for cross-coroutine longjmp

Handlers using `setjmp` can lead a second handler on the inside to jump from one coroutine to the next, which leads `-fsanitize=address` to fail.

Inform minicore about each `setjmp` and `longjmp` so that stays in sync. (PR: https://github.com/jfecher/minicoro/pull/2)

Also includes a test case that fails even without `-fsanitize=address`

Note: the submodule commit ID here needs to be updated after the minicore squash merge

### DIFF
--- a/examples/codegen/effects/abort_within_coroutine.an
+++ b/examples/codegen/effects/abort_within_coroutine.an
@@ -1,0 +1,49 @@
+// The setjmp side of the abort handler lives inside a coroutine: the
+// outer foldr's generator runs on a coroutine stack, and the `try`
+// performed there is aborted from a second, nested coroutine. After the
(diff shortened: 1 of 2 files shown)


jfecher:
From the discussion in the linked PR:

> I'm currently leaning towards:
> - Disable the 0-resumptive effect -> abort optimization
> - Add the test in the parent PR as a regression test

Do you mind editing this PR so it is just the regression test and disabling the abort optimization?
Otherwise I can make the change myself once I finish the trait/effect split work.

lichtblau:
Sure thing: https://github.com/jfecher/ante/pull/256

PR #255, opened 2026-07-21
Share compiler options across build and run

rupert648:
This MR tries to tidy up the CLI flags//struct a bit as part of #235 .

This makes flags like --backend and --release work consistently with both subcommands, and lets ante run pass arguments through to the program i.e. allowing a program to take in its args


jfecher (approved):
LGTM, thank you!

Appreciate the addition of run args

PR #256, opened 2026-07-23
Test case for cross-coroutine longjmp

lichtblau:
Test case for cross-coroutine longjmp


jfecher:
This gives a type error (once changed to use traits/effects) with the traits/effects split now. I'm going to see if I can adapt it though to still get the original crash.

jfecher:
Ty for the investigation! Fixed in d0af88d50a978a89bc85899f2787a97cd71f4804

PR #257, opened 2026-07-26
cleanup error handling and add an `init` command

rupert648:
Cleans up error handling to surface all errors to the runtime. We use system exit (1) so we can get a pretty message, we introduce thiserror so we can pass errors up the chain rather than scattering system exits everywhere.

Introduces a basic init command which inits a project at the given path (defaults to `.`). This creates an `ante.toml` and `src/main.an` with a hello world example. It tries to error out early where applicable. Example

```
~/Documents/personal/ante on  master ✗ ⌚ 15:21:49
$ ./target/debug/ante init ../hello-world

~/Documents/personal/ante on  master ✗ ⌚ 15:22:05
$ cat ../hello-world/ante.toml
name = "hello-world"

~/Documents/personal/ante on  master ✗ ⌚ 15:22:19
$ cat ../hello-world/src/main.an
main () = println "Hello, World!"

~/Documents/personal/ante on  master ✗ ⌚ 15:22:39
$ cd ../hello-world

~/Documents/personal/hello-world ⌚ 15:23:06
$ ../ante/target/debug/ante run
Hello, World!

~/Documents/personal/hello-world ⌚ 15:23:19
$ cd ../ante

~/Documents/personal/ante on  master ✗ ⌚ 15:23:23
$ ./target/debug/ante init ../hello-world
error: Failed to init project: Manifest File Already Exists

~/Documents/personal/ante on  master ✗ ⌚ 15:23:29
$
```

Also ran `cargo clippy --fix`

Let me know what you think of the use of `thiserror` and the current setup for init - a lot of this stuff is preference (e.g. over anyhow) so v happy if you have a preferred way of doing things

### DIFF
--- a/src/dependencies.rs
+++ b/src/dependencies.rs
@@ -14,7 +14,7 @@ pub enum DependencyError {
 impl fmt::Display for DependencyError {
     fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Self::MissingProjectRoot => write!(formatter, "Missing project root; please create one with `ante init`"),
+            Self::MissingProjectRoot => write!(formatter, "Missing project root, please create one with `ante init`"),
             Self::Io(error) => error.fmt(formatter),
             Self::GitClone(error) => write!(formatter, "Git clone failed: {error}"),
         }
--- a/src/dependencies.rs
+++ b/src/dependencies.rs
@@ -1,38 +1,60 @@
-use std::process::Command;
+use std::{error::Error, fmt, io::Error as IoError, path::PathBuf, process::Command};
 
-use colored::Colorize;
-
-use crate::{find_files::find_nearest_project_root, manifest::MANIFEST_FILE_NAME};
+use crate::find_files::find_nearest_project_root;
 
 pub const DEPENDENCIES_DIR_NAME: &str = "deps";
 
-/// Add a git dependency to the current Ante project by cloning it into the dependencies directory.
-pub fn add_git_dependency(dep_url: &str) {
-    let Some(project_root) = find_nearest_project_root() else {
-        eprintln!(
-            "{}: could not find {MANIFEST_FILE_NAME} in the current directory or any parent directory",
-            "error".red(),
-        );
-        std::process::exit(1);
-    };
+#[derive(Debug)]
+pub enum DependencyError {
+    MissingProjectRoot,
+    Io(IoError),
+    GitClone(String),
+}
 
-    let deps_dir = project_root.join(DEPENDENCIES_DIR_NAME);
-    if let Err(error) = std::fs::create_dir_all(&deps_dir) {
-        eprintln!("{}: failed to create `{}`:\n{error}", "error".red(), deps_dir.display());
-        std::process::exit(1);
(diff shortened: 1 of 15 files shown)


src/dependencies.rs
```diff
@@ -1,38 +1,44 @@
...
-use colored::Colorize;
-
-use crate::{find_files::find_nearest_project_root, manifest::MANIFEST_FILE_NAME};
+use crate::find_files::find_nearest_project_root;
 
 pub const DEPENDENCIES_DIR_NAME: &str = "deps";
 
+#[derive(Error, Debug)]
+pub enum DependencyError {
+    // TODO: implement ante init
```
jfecher:
```suggestion
```

src/dependencies.rs
```diff
@@ -1,38 +1,44 @@
...
+    #[error("Missing Project Root; please create one with `ante init`")]
```
jfecher:
```suggestion
    #[error("Missing project root: please create one with `ante init`")]
```

src/main.rs
```diff
@@ -129,6 +130,14 @@ impl CompileRequest {
 }
 
 fn main() {
+    // wrapper so we can nicely print command errors
```
jfecher:
```suggestion
```
I don't think the comment is necessary

src/project.rs
```diff
@@ -0,0 +1,128 @@
...
+use std::{
+    io::Error as IoError,
+    path::{Path, PathBuf},
+};
+use thiserror::Error;
+
+/// Default name of the src folder.
+pub(crate) const SRC_FOLDER: &str = "src";
+pub(crate) const MAIN_FILE: &str = "main.an";
+const DEFAULT_MAIN_SOURCE: &str = include_str!("../templates/default_main.an");
```
jfecher: I'd somewhat prefer for this string to be inlined to avoid the extra templates directory for now. It can be re-added if the compiler ever starts using many more kinds of templates.

Let's also change the template to indent the body of main:
```hs
main () =
    println "Hello, World!"
```

jfecher (commented):
Looks good, just a couple of suggestions.

Do you mind also removing `thiserror`? I'd just like to err on the side of fewer dependencies for now

src/project.rs
rupert648: done! Also added a little regression test to ensure we're always generating a valid program

rupert648:
removed thiserror

src/dependencies.rs
```diff
@@ -1,38 +1,60 @@
...
+pub enum DependencyError {
+    MissingProjectRoot,
+    Io(IoError),
+    GitClone(String),
+}
+
+impl fmt::Display for DependencyError {
+    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::MissingProjectRoot => write!(formatter, "Missing project root; please create one with `ante init`"),
```
jfecher:
```suggestion
            Self::MissingProjectRoot => write!(formatter, "Missing project root, please create one with `ante init`"),
```

jfecher (dismissed):
LGTM, thank you!

PR #258, opened 2026-07-26
fix: update ante-ls to reference effect/trait instead of ability

rupert648:


ante-ls/src/exports.rs
```diff
@@ -64,7 +58,19 @@ fn find_top_level_name_at(
...
+                searcher
+                    .try_offer(loc.span.start.byte_index, loc.span.end.byte_index)
+                    .then(|| ctx.get_name(def.name).to_string())
+            },
+            TopLevelItemKind::EffectDefinition(def) => {
+                let loc = &ctx.name_locations[def.name];
+                searcher
+                    .try_offer(loc.span.start.byte_index, loc.span.end.byte_index)
+                    .then(|| ctx.get_name(def.name).to_string())
+            },
```
jfecher:
```suggestion
            TopLevelItemKind::TraitDefinition(def) | TopLevelItemKind::EffectDefinition(def) => {
                let loc = &ctx.name_locations[def.name];
                searcher
                    .try_offer(loc.span.start.byte_index, loc.span.end.byte_index)
                    .then(|| ctx.get_name(def.name).to_string())
            },
```

jfecher (dismissed):
Thank you! I should have caught this earlier.

jfecher:
@rupert648 since I reviewed this while you were still making changes, let me know once changes are done and I'll re-review if necessary and merge

ante-ls/src/exports.rs
rupert648: oh lol ofc, fixed

jfecher (approved):

PR #261, opened 2026-08-05
fix: hashmap bug across collisions and tombstones

rupert648:
This change fixes a bug in key collisions/deletions

Take A and B where there is a collision
[n] -> A
[n+1] -> B
when A is deleted
[n] -> tombstone
[n+1] -> B
if we don't continue iterating to search for B when we see a tombstone we would get [n] -> B
[n+1] -> B
if we re-inserted B

to fix this we continue probing past collision entries during insertion to avoid creating duplicate keys. Remembering the tombstone preserves the best insertion location if the key isn't found elsewhere.

We also add some regression coverage.

### DIFF
--- a/examples/codegen/stdlib-tests/map.an
+++ b/examples/codegen/stdlib-tests/map.an
@@ -100,7 +100,7 @@ test_collision_management () =
     println (m.get first ~> try)
     println (m.get second ~> panic_on_fail)
     println (m.insert second 3)
-    println (m.get (ref second) ~> panic_on_fail)
+    println (m.get second ~> panic_on_fail)
     println (m.len ())
 
 
--- a/examples/codegen/stdlib-tests/map.an
+++ b/examples/codegen/stdlib-tests/map.an
@@ -97,8 +97,8 @@ test_collision_management () =
     // testing tombstone case
     m.remove (ref first)
     println (m.len ())
-    println (m.get (ref first) ~> try)
-    println (m.get (ref second) ~> panic_on_fail)
+    println (m.get first ~> try)
+    println (m.get second ~> panic_on_fail)
     println (m.insert second 3)
     println (m.get (ref second) ~> panic_on_fail)
     println (m.len ())
--- a/examples/codegen/stdlib-tests/map.an
+++ b/examples/codegen/stdlib-tests/map.an
@@ -91,8 +91,8 @@ test_collision_management () =
     // these have the same hash so should be put in the same index
     println (m.len ())
 
-    println (m.get (ref first) ~> panic_on_fail)
-    println (m.get (ref second) ~> panic_on_fail)
+    println (m.get first ~> panic_on_fail)
+    println (m.get second ~> panic_on_fail)
 
     // testing tombstone case
     m.remove (ref first)
--- a/examples/codegen/stdlib-tests/map.an
+++ b/examples/codegen/stdlib-tests/map.an
@@ -2,7 +2,7 @@ import Std.HashMap.HashMap, empty, of, clear, insert, print_hashmap, remove, get
     get_entry, stream_map
 import Std.Stream.iter, iota, map, stream_fn, stream_array
 import Std.Hash.Hash, hash_u8
-import Std.Fail.try
+import Std.Fail.try, panic_on_fail
 
 main () =
     var map = empty ()
@@ -32,6 +32,7 @@ main () =
     test_iter_and_clear ()
     test_get_or_insert_contains_and_merge ()
     test_insert_duplicate ()
+    test_collision_management ()
 
 test_iter_and_clear () =
     // stream_map impl: build a single-entry map (otherwise ordering isn't deterministic)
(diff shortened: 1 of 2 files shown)


examples/codegen/stdlib-tests/map.an
```diff
@@ -69,6 +70,40 @@ test_insert_duplicate () =
...
+test_collision_management () =
+    first = MyType 1
+    second = MyType 2
+        
+    var m = HashMap.of [(first, 1), (second, 2)]
+    // these have the same hash so should be put in the same index
+    println (m.len ())
+
+    println (m.get (ref first) ~> panic_on_fail)
+    println (m.get (ref second) ~> panic_on_fail)
```
jfecher:
```suggestion
    println (m.get first ~> panic_on_fail)
    println (m.get second ~> panic_on_fail)
```
These explicit refs are unnecessary

examples/codegen/stdlib-tests/map.an
```diff
@@ -69,6 +70,40 @@ test_insert_duplicate () =
...
+
+    // testing tombstone case
+    m.remove (ref first)
+    println (m.len ())
+    println (m.get (ref first) ~> try)
+    println (m.get (ref second) ~> panic_on_fail)
```
jfecher:
```suggestion
    println (m.get first ~> try)
    println (m.get second ~> panic_on_fail)
```

examples/codegen/stdlib-tests/map.an
```diff
@@ -69,6 +70,40 @@ test_insert_duplicate () =
...
+    println (m.insert second 3)
+    println (m.get (ref second) ~> panic_on_fail)
```
jfecher:
```suggestion
    println (m.get second ~> panic_on_fail)
```

stdlib/src/HashMap.an
```diff
@@ -73,7 +73,8 @@ HashMap.insert (map: mut HashMap k v) (key: k) (value: v) {Hash k} {Eq k}: Maybe
         map.resize ((map.capacity + 1) * 2)
 
     var previous = Maybe.None
-    iter_until (map: mut HashMap k v) (key: ref k) (value: ref v) (start: Usz) (end: Usz) : Bool =
+    var first_tombstone: Maybe Usz = None
```
jfecher:
```suggestion
    var first_tombstone = Maybe.None
```

stdlib/src/HashMap.an
```diff
@@ -85,16 +86,31 @@ HashMap.insert (map: mut HashMap k v) (key: k) (value: v) {Hash k} {Eq k}: Maybe
...
+            else if entry.tombstone then
+                if first_tombstone is None then
+                    first_tombstone := Some start
+
+                iter_until map key value (start + 1) end
             else
-                entry := Entry (deref key) (deref value) true false
+                insertion_index = match first_tombstone
+                    | Some index -> index
+                    | None -> start
```
jfecher:
```suggestion
                insertion_index = first_tombstone.unwrap_or_else do start
```

stdlib/src/HashMap.an
```diff
@@ -85,16 +86,31 @@ HashMap.insert (map: mut HashMap k v) (key: k) (value: v) {Hash k} {Eq k}: Maybe
...
 
     h = Usz (Hash.hash (deref (ref key))) % map.capacity
 
     if not iter_until map key value h map.capacity then
         if not iter_until map key value 0 h then
-            panic "Failed to insert entry into map"
+            match first_tombstone
+            | Some index ->
+                entry = ptr_to_mut (offset map.entries index)
+                entry := Entry (deref key) (deref value) true false
```
jfecher:
```suggestion
                array_insert map.entries index (Entry (deref key) (deref value) true false)
```

stdlib/src/HashMap.an
```diff
@@ -121,8 +137,11 @@ HashMap.get_entry_mut (map: mut HashMap k v) (key: ref k) {Hash k} {eq: Eq k}: M
         else
             entry_ptr = offset map.entries start
             entry = deref_ptr entry_ptr
-            if entry.occupied and ref entry.key == key then
-                Found entry_ptr
+            if entry.occupied then
+                if ref entry.key == key then
```
jfecher:
```suggestion
                if entry.key == key then
```
I don't think this should be necessary either but sometimes traits mess things up

examples/codegen/stdlib-tests/map.an
```diff
@@ -69,6 +70,40 @@ test_insert_duplicate () =
...
+    inner: Usz
+
+impl copy_mytype: Copy MyType with
+      (.*) = deref
+
+impl hash_mytype: Hash MyType with
+    hash _ = 1
+
+impl equal_mytype: Eq MyType with
+    (==) x y = x.inner == y.inner
```
jfecher, line 84: May be simpler to just create your own hash impl for Usz since Ante has no global coherence rule:
```rs
impl hash_to_one: Hash Usz with
    hash _ = 1
```

jfecher (commented):
The Ante code here looks a bit AI-generated, please disclose any AI use in the future, I'll probably add this to a CONTRIBUTING.md file but it isn't blocking.

Thanks for the PR! This same issue happened in another HashMap impl I wrote for a different language/compiler and I suppose I forgot to update Ante as well - thanks for finding this.

jfecher (approved):

rupert648:
> The Ante code here looks a bit AI-generated, please disclose any AI use in the future, I'll probably add this to a CONTRIBUTING.md file but it isn't blocking.
> 
> Thanks for the PR! This same issue happened in another HashMap impl I wrote for a different language/compiler and I suppose I forgot to update Ante as well - thanks for finding this.

Hey! This was hand written :) I use AI a lot in my work day and working on Ante bits has been my break from that to keep my mind active lol. Would be interested to hear what looked AI generated about it? I spotted this bug as I have been writing a HashSet implementation in Ante and was using Hashmap as the reference point

jfecher:
Apologies for the accusation then! For what its worth I never suspected anything in your previous PRs, some of the Ante code in this one just caught my eye a bit. From what I've seen, AI tends to overuse `ref` in Ante, uses explicit type annotations where they are unnecessary, uses explicit matches a lot, and seems to avoid the `do` keyword.

Always happy to see your PRs though, thank you for all the contributions! Is the HashSet part of a larger project?

rupert648:
> Is the HashSet part of a larger project?

Not really! Just been writing random small bits of a Ante to get used to things/learn the language, HashSet is a gap in the stdlib so thought it would be a nice one to try and requires a bit more use of antes features. Although that said 90% of the code will just be ripped from the HashMap implementation

jfecher:
> Although that said 90% of the code will just be ripped from the HashMap implementation

It would be nice if the HashSet could hold a `HashMap k Unit` internally and defer most methods to that, similar to what Rust does. Unit values aren't 0-sized in Ante yet but they are planned to be in the future so it should be just as efficient.

Issue #263, opened 2026-08-10
Mismatch between MIR zero sized unit type and C backend output


rupert648:
Reasonably simple - MIR marks Unit sizes as having a size of `0`:
https://github.com/jfecher/ante/blob/master/src/mir/mod.rs#L1069

This works fine for LLVM, where Unit is lowered into an empty LLVM struct:
https://github.com/jfecher/ante/blob/master/src/codegen/llvm/mod.rs#L440

However for C, the output for a Unit is a struct with an _unused in it
https://github.com/jfecher/ante/blob/master/src/codegen/c/cfile.rs#L150

The sizeof this struct is 1.

## Actual Output

This results in incorrect pointer arithmetic which is trivial to show, e.g. iterating a Vec of structs with Unit fields:
```
import Std.Vec.Vec, stream_imm_vec
import Std.Stream.iter, stream_array

type MyType =
    key: U8
    idk: Unit
    aboolean: Bool

print_mytype (res: imm MyType) =
    print res.key
    print ", "
    println res.aboolean

main () =
    x = Vec.of [MyType 1 () true, MyType 2 () false]
    iter (imm x) print_mytype
```

```
~/Documents/personal/ante-unit-zero ⌚ 20:50:42
$ ../ante/target/debug/ante --backend c src/main.an
1, true
1, false

~/Documents/personal/ante-unit-zero ⌚ 20:51:06
$ ../ante/target/debug/ante  src/main.an
1, true
2, false
```
## Expected Output

C backend output should match llvm output, and produce correct results

## Possible Solutions
The simplest solution is to just set MIR PrimitiveType::Unit size to 1 instead of zero and change the llvm output type to match e.g. (or could just directly lower to i8)
```
PrimitiveType::Unit => self.llvm.struct_type(&[self.llvm.i8_type().into()], false).into()
```
As ante wants to work towards zero sized abstractions/types tho. The better solution is probably to fix forward C to omit Unit types from *storage*, e.g. tuples, structs etc.

i.e. maintain the original Unit struct for standalone values, but omit it from struct representations
```
typedef struct {
      uint8_t _0;
      Unit _1; 
      bool _2;
      bool _3;
  } Entry;
```
would now be
```
ypedef struct {
      uint8_t _0;
      // Logical field _1 is Unit and requires no storage
      bool _2;
      bool _3;
  } Entry;
```
[...]

Issue #238, opened 2026-05-29
Filter out zero-sized types in Mir


jfecher:
Zero-sized types like `Unit` are invalid C so they are instead represented as a single byte struct in the C backend which can lead to different behavior from the llvm backend. To fix this there could be a Mir pass to automatically filter out zero-sized types out of argument lists & types.

PR #264, opened 2026-08-11
Fix Unit layout consistency across backends

rupert648:
This changes the two zero sized types ante has; empty tuples and Unit. To both be 1 sized in the llvm backend (and marked as 1 sized in MIR pass). See #263

Whilst the longer term goal (#238 ) is to be zero sized everywhere, this requires more design/a longer term approach. Right now the sizing is inconsistent between the C backend (which produces a 1 sized struct) and the LLVM backend which produces a zero sized struct. The former causes pointer arithmetic bugs as can be seen in the regression tests in this MR or in #263 .

For now, until zero sized types are implemented for the C backend (or during the MIR passthrough as aforementioned in #238), we take the simpler route to make things consistent by setting the size of each to be 1.

### DIFF
--- a/examples/codegen/unit_pointer_offset.an
+++ b/examples/codegen/unit_pointer_offset.an
@@ -1,0 +1,28 @@
+import Std.Vec.Vec, stream_imm_vec
+import Std.Stream.iter, stream_array
+
+type MyType =
+    key: U8
+    idk: Unit
+    aboolean: Bool
+    tombstone: Bool
+
+print_mytype (res: imm MyType) =
+    print res.key
+    print ", "
+    println res.aboolean
+
+main () =
+    x = Vec.of [MyType 1 () true true, MyType 2 () false false]
+    iter (imm x) print_mytype
+
+
+    println (size_of (MkType: Type Unit))
+    println (size_of (MkType: Type MyType))
+
+// args: --delete-binary --no-color
+// expected stdout:
+// 1, true
+// 2, false
+// 1
+// 4
--- a/src/codegen/llvm/mod.rs
+++ b/src/codegen/llvm/mod.rs
@@ -7,7 +7,7 @@ use inkwell::{
     module::{Linkage, Module},
     passes::PassBuilderOptions,
     targets::{CodeModel, FileType, InitializationConfig, RelocMode, Target, TargetMachine},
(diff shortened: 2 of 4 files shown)


examples/codegen/unit_pointer_offset.an
```diff
@@ -0,0 +1,21 @@
...
+    aboolean: Bool
+
+print_mytype (res: imm MyType) =
+    print res.key
+    print ", "
+    println res.aboolean
+
+main () =
+    x = Vec.of [MyType 1 () true, MyType 2 () false]
+    iter (imm x) print_mytype
```
jfecher, line 17: Can you combine this test with unit_layout.an?

They are related and both small enough to be merged

src/codegen/llvm/mod.rs
```diff
@@ -526,8 +528,16 @@ impl<'ctx> ModuleContext<'ctx> {
...
+    fn empty_product_type(&self) -> StructType<'ctx> {
+        self.llvm.struct_type(&[self.llvm.i8_type().into()], false)
+    }
+
+    fn empty_product_value(&self) -> BasicValueEnum<'ctx> {
+        self.empty_product_type().const_zero().into()
+    }
+
+    fn unit_value(&self) -> BasicValueEnum<'ctx> {
+        self.empty_product_value()
```
jfecher: I think we can merge the notion of an empty product type and the unit type and just use one `unit_value` & type method. Unit is often defined as the empty tuple and the `Type::Tuple` type is mostly an internal detail for the compiler to more easily lower closures anyway. Actual tuples in ante code use the pair type which would be a `Type::Application(Type::UserDefined(pair_type_id), vec![a, b])`. So I don't think it's important we distinguish between the two.

jfecher (commented):
Looks good! Just some minor code organization suggestions

src/codegen/llvm/mod.rs
rupert648: sounds good, made it just `unit_value` and `unit_type`

jfecher (approved):
LGTM!

jfecher:
No need to force push btw. PRs should be squashed on merge automatically

rupert648:
> No need to force push btw. PRs should be squashed on merge automatically

force of habit, we don't squash where I work so I like keeping a clean history

PR #265, opened 2026-08-11
stdlib: take inequality operands by reference

rupert648:
As said on tin, all implementations of == take references (see `trait Eq t`)
Therefore it should be safe to also change neq to take this too. Otherwise code like below fails to compile with a value moved error:
```ante
func (left: s) (right: s) {Eq s} =
    if left != right then
        println left
        println right
```


jfecher (approved):
Good catch. I must have just missed this when making the original change.

I'm still unsure if the stdlib should prioritize `ref t` or `t` + `Copy t` in cases where the argument shouldn't be consumed, but best to be consistent here.

PR #266, opened 2026-08-11
fix(#269): matching Unit variant payloads

rupert648:
pattern matching a unit type in a match expression e.g.
```ante
match result
| Ok () -> println "ok"
| Error error -> println error
```
resulted in the following panic when compiling:
```
thread '<unnamed>' panicked at src/mir/validation.rs:
  Switch value type is not an integer, got `Unit`
```
digging in, I think this is because MIR generates effectively a nested match statement for the `Ok ()` pattern matching. The "discriminant" of this nested switch is of type Unit, but it creates a switch with cases of integer, something like:
```
switch Unit
| 0 -> ...
```
which causes the panic.

The fix should be to simply represent Unit in these cases as `0`
so instead we get something like
```
switch 0
| 0 -> ...
```
Again this would probably be resolved in the longer term fix in #238
to filter out zero sized types.

### DIFF
--- a/examples/regressions/269_match_unit_payload.an
+++ b/examples/regressions/269_match_unit_payload.an
@@ -1,0 +1,9 @@
+main () =
+    result: Result Unit String = Ok ()
+
+    match result
+    | Ok () -> println "ok"
+    | Error error -> println error
+
+// args: --delete-binary --no-color
+// expected stdout: ok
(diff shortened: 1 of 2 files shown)


examples/regressions/match_unit_payload.an
rupert648, line 1: should I create an issue for this problem so regressions are properly tagged?

jfecher: Yes, an issue would be nice!

examples/regressions/269_match_unit_payload.an
rupert648: done! #269

jfecher (approved):
Thank you! Good investigation as well

PR #268, opened 2026-08-14
Add float-to-string casts (#267)

### LINKED ISSUE #267: Add `Cast F32 String` and `Cast F64 String`
The prelude has impls for `Print F32` and `Print F64` but it seems I forgot to add the corresponding impls to cast floats to strings, making floats unusable in string interpolation `"foo ${my_float} bar"`. We should add these impls in the prelude.

floze-the-genius:
## Summary

- add exported `Cast F32 String` and `Cast F64 String` implementations
- keep their output consistent with the existing `print_float` formatting
- cover both float types through the string-interpolation golden test

Closes #267.

## Testing

- `cargo test --no-default-features --test goldentests -- --nocapture` — 198/198 primary-backend cases and 77/77 C-backend cases passed
- `cargo test --no-default-features --workspace --lib --bins` — 24 + 38 + 13 unit tests passed
- `cargo test --no-default-features --test cli` — 1/1 passed
- direct C-backend regression: `f32=1.2500, f64=-2.5000`

LLVM 21 is not installed locally, so the tests used `--no-default-features`; the repository CI will exercise the default LLVM build. `cargo fmt --check` and strict `cargo clippy -D warnings` also report existing findings only in untouched Rust files; this PR changes two `.an` files.

### DIFF
--- a/examples/codegen/string_interpolation.an
+++ b/examples/codegen/string_interpolation.an
@@ -2,7 +2,11 @@ main () =
     name = "world"
     n: I32 = 42
     println "hello, ${name}! n=${n}"
+    f32: F32 = 1.25f
+    f64: F64 = -2.5
+    println "f32=${f32}, f64=${f64}"
 
 // args: --delete-binary --no-color
-// expected stdout: hello, world! n=42
-
+// expected stdout:
+// hello, world! n=42
+// f32=1.2500, f64=-2.5000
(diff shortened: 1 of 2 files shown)


jfecher (approved):
Thank you - please avoid AI-generated PR descriptions in the future though

PR #272, opened 2026-09-09
Add HashSet implementation

rupert648:


jfecher (approved):
LGTM!

I'm going to be reworking the stdlib tests to use asserts over printlns so the expected output is closer to the source. I'll make that as a separate change after this is merged though, along with some other minor adjustments. Thanks for the PR!

